PR #4, opened 2016-01-21
Some cloudimage fixes

fredrikt:
These are low risk changes, I don't expect anything to break. However, since I don't know if sunet::cloudimage is in use somewhere critical, I want to coordinate inclusion in master.


leifj:
looks good

PR #5, opened 2019-10-23
White list docker compose file format 3.7

glance-:
Our new sunet-jenkins setup uses docker compose file format 3.7, and it
still works as check_docker_containers expects, so white list that
version to.


fredrikt:
merged on the sunet_dev branch

PR #7, opened 2019-11-25
Docker registry read only clean

glance-:
This starts to use new functionality in docker-registry-auth to put the service in read-only while cleaning it, instead of stopping the service.


glance-:
Next phase after this is:
https://github.com/glance-/puppet-sunet/compare/docker-registry-read-only-clean...glance-:docker-registry-rework
Which moves the registry services from compose to docker-run straight under systemd to reduce the dependency footprint.

After that I've prepared some registry-cleaning-rework, by stop removing bits in the filesystem layer directly and rather DELETE them over https and then let the garbage collector clean it up:
https://github.com/glance-/puppet-sunet/compare/glance-:docker-registry-rework..glance-:docker-clean_registry-rework

manifests/docker_registry.pp
```diff
@@ -39,15 +38,15 @@
 
     package {['python-yaml', 'python-ipaddress', 'python-requests']:
         ensure => installed
-    } ->
+    }
 
-    file { "${registry_cleanup_basedir}/clean_registry_cron":
+    -> file { "${registry_cleanup_basedir}/clean_registry_cron":
```
fredrikt: We don't use -> to order execution anymore, we have "ordering = manifest" in puppet.conf. That was a "recommended and safe" new setting in Puppet 3.8, but the link to that page is broken now.

I know you didn't introduce them and only reformatted them, but please remove them rather than reformatting them.

PR #8, opened 2019-11-26
Docker registry rework

glance-:
Move registry from sunet::docker_compose to docker_run

These are run as 3 different containers, and not in a compose group
to simplify their running and ensure they keep running in case of
system degradation and/or cleanup scripts going berserk.

This way, we can control the startup order better, and have the
registry-auth container come out of the "backdoor" of the registry.
    
This way we can be independent of any images in local cache.


manifests/docker_run.pp
```diff
@@ -40,18 +49,16 @@
   }
 
   $image_tag = "${image}:${imagetag}"
-  docker::image { "${name}_${image_tag}" :  # make it possible to use the same docker image more than once on a node
-    image   => $image_tag,
-    require => Class['sunet::dockerhost'],
+  if $docker_image {
```
fredrikt: The name $docker_image is misleading, but rather than changing that why don't you use ensure_resource('Docker::Image', ...) instead?

glance-: Whats the difference between the current construct and running ensure_resource?

We can't run docker::image for localhost:5000/sunet/docker-registry-auth , when the registry container isn't up yet, and that happens towards the end of the puppet run.

run_docker_image ? fetc_docker_image ? or some other better idea?

glance-: ensure_resource can avoid conflicting/multiple resource declarations, but that's not the issue here. We just need to turn off the call to docker::image.

fredrikt: ok, then don't call the argument $docker_image since that name is commonly used as the name of the docker image

glance-: Ok, renamed it to fetch_docker_image.

PR #13, opened 2021-12-03
Include ufw in sunet::misc::ufw_allow

ernstwi:
Would it make sense to include ufw from this file? I ran into this when setting up a new machine with only `sunet::dockerhost` which uses `ufw_allow` but does not `include ufw`. Other classes like `sunet::server` which call `ufw_allow` does include ufw explicitly.

Maybe there is a more elegant solution than putting the include statement inside the loop, but it is not a technical problem to have repeated includes, per the docs.


fredrikt:
outdated

PR #20, opened 2023-02-02
Kano merge drive

mickenordin:
This patchset merges Drive and Main branch. It contains a few things not all ready in main, like:

* support for setting owner and  permissions to compose dir
* fix a problem if instance name is longer that 12 chars, an interface can not be create in loadbalancer
* using a common template for all loadbalancer instances
* adding more config options for updater
* allowing setting more settings in docker-compose for frontend like dns and exposed ports

All changes should be backwards compatible with existing stuff


theseal:
👍

PR #21, opened 2023-02-07
add support for iif parameter for docker_expose.nft 

indy-independence:
add support for iif parameter for docker_expose.nft so you can change the interface you want to publish services on, in safespring you might need to publish on wireguard overlay tunnel interface for example


mickenordin:
This is backwards compatible and seems safe, so I am merging it :)

PR #26, opened 2023-02-15
Make docker containers able to reach outside via any interface, not just via eth0

indy-independence:
When having multiple interafces on the host like eth0 and wg0 this will allow containers to reach things behind any of those interfaces by omitting eth0 in rules and just filter on to_docker interface instead.

Tested on a host with eth0 and wg0, conntrack states are added correctly both for outgoing traffic via eth0 and wg0. Traffic to docker host is not affected and still sees to_docker ip addresses only.

### DIFF
--- a/templates/dockerhost/200-dockerhost_nftables.nft.erb
+++ b/templates/dockerhost/200-dockerhost_nftables.nft.erb
@@ -18,18 +18,18 @@ table ip6 nat {
 #
 # SNAT packages _from_ Docker. Can't use iifname in postrouting rules.
 #
-add rule ip nat postrouting ip saddr { 172.16.0.0/12 } oif eth0 counter masquerade comment "SNAT traffic from Docker"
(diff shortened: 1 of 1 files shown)


indy-independence:
nft postrouting, traffic counter increases:
```
	chain postrouting {
		type nat hook postrouting priority srcnat; policy accept;
		ip saddr 172.16.0.0/12 iif "to_docker" counter packets 4 bytes 336 masquerade comment "SNAT traffic from Docker"
	}
```
conntrack:
```
# conntrack -L
icmp     1 27 src=172.16.0.2 dst=10.102.0.3 type=8 code=0 id=15 src=10.102.0.3 dst=10.102.0.5 type=0 code=0 id=15 mark=0 use=1
icmp     1 24 src=172.16.0.2 dst=172.16.0.1 type=8 code=0 id=14 src=172.16.0.1 dst=172.16.0.2 type=0 code=0 id=14 mark=0 use=1

```

PR #29, opened 2023-02-17
better nftables fact

fredrikt:


mickenordin:
Tested on debian 11 and it works, so merging

PR #30, opened 2023-02-21
Sunet frontends 2.0

fredrikt:
squash-merge 11 months of Sunet Frontend 2.0 development from the eduid_dev branch
    
    Noteworthy changes:
    
      - nothing runs as uid/gid 0
      - uses nftables
      - working IPv6 connectivity to _and_ from haproxy containers
      - update to Python 3 in various helper scripts
      - exabgp runs outside of Docker (because of TCP-MD5 and --net host not
        working in namespaced Docker)


mickenordin:
We did a full code review today, and it seems backwards compatible, so I will merge this. When I have time I will also test it in testing branch. Note: eduid has frontend running this code allready, from eduid branch of course

PR #31, opened 2023-02-23
avoid starting services and setting up firewall rules before docker-in-namespace is ready

fredrikt:
- We need to set up the /etc/systemd/system/docker.service.d/ docker_nftables_ns.conf file before the first start of Docker on a new
  host, to not require a reboot before Docker actually works with the
  new to_docker interface (the old docker0 interface interferes).

- We can't add firewall rules referencing the interface "to_docker" before it actually exists (since that will make nftables not load the rules at all).

### DIFF
--- a/manifests/dockerhost.pp
+++ b/manifests/dockerhost.pp
@@ -21,6 +21,40 @@ class sunet::dockerhost(
   include sunet::packages::python3_yaml # check_docker_containers requirement
   include stdlib
 
+  if $::facts['sunet_nftables_enabled'] == 'yes' {
+    # Hackishly create the /etc/systemd/system/docker.service.d/ directory before the docker service is installed.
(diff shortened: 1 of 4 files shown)


berrabou:
Use iifname and oifname in nftables if you want to reference interfaces before they exists.

manifests/dockerhost.pp
```diff
@@ -21,6 +21,40 @@
   include sunet::packages::python3_yaml # check_docker_containers requirement
   include stdlib
 
+  if $::facts['sunet_nftables_enabled'] == 'yes' {
+    # Hackishly create the /etc/systemd/system/docker.service.d/ directory before the docker service is installed.
```
theseal, line 25: This is where my attempted ended yesterday 😁

theseal (approved):
LGTM

PR #33, opened 2023-03-01
Don't throw away v6 route when IP forwarding

theseal:
In environments where we relay on RAs for IPv6 (e.g Safespring) we need to forcely allow RAs even if IP forwarding is enabled.

We can't use 'all' when configuring "accept_ra" so specify the primary interface.
https://unix.stackexchange.com/questions/90443/what-is-the-difference-between-all-default-and-eth-in-proc-sys-net-ipv


theseal:
Solved in https://github.com/SUNET/multiverse/pull/38

PR #38, opened 2023-03-17
Unbound: Break out disable_resolved_stub to it's own class

mickenordin:
We might need to disable stub resolver in other cases than for unbound so this patch makes a separate class for this, which is then used in the unbound class, making this patch backwards compatible with the old use case.


theseal:
Seems to work just fine 👍

PR #42, opened 2023-04-04
Fixes for naemon

theseal:
Will backport/cherry-pick 94edf559a313636788eec6316e306efa485893b5 to the legacy sunet branch. All other changes are for stable and nftables only.


theseal:
Wrapped for legacy OS. Please have another look.

PR #50, opened 2023-05-05
Merge satosa classes from ops repos

theseal:
Merge satosa classes from "all" ops repos so that we only need to maintain one class.

Can't find any ops repo that uses this class (`sunet::satosa`) so it should be safe to not be backwards compatible.

Some tweaks may be needed `cdb.json` when the class is used against a real Open Id Connect RP.

### DIFF
--- a/manifests/satosa.pp
+++ b/manifests/satosa.pp
@@ -65,7 +65,7 @@ class sunet::satosa(
   }
 
   if ($dehydrated_name) {
-    sunet::dehydrated::client { 'enable-dehydrated':
+    class { 'sunet::dehydrated::client':
       domain    => $dehydrated_name,
       ssl_links => true,
     }
--- a/manifests/satosa.pp
+++ b/manifests/satosa.pp
@@ -50,6 +50,16 @@ class sunet::satosa(
     }
   }
 
+  $json_config = hiera('satosa_json_config')
+  sort(keys($plugins)).each |$n| {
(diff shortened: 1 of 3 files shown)


manifests/satosa.pp
```diff
@@ -1,79 +1,121 @@
...
+      cert_file => '/etc/satosa/https.crt'
+    }
+  }
+
+  if ($enable_oidc) {
+    $client_id     = lookup('client_id')
+    $client_secret = lookup('client_secret')
+    file { '/etc/satosa/cdb.json':
+      ensure  => file,
+      content => template('sunet/satosa/cdb.json.erb')
```
mickenordin, line 102: Should we make it possible to add your own cdb.json here? Possibly by adding a "manage_cdb" parameter to the class?

mickenordin (changes requested):
Maybe my comment about being able to supply your own cdb.json should be a thing?

manifests/satosa.pp
theseal: I will try to completely rewrite this segment with you comment in mind. Stay tuned.

theseal: A few commits later and the json configurations are now handled in the same way as all other yaml configuration in this class. Think it got a lot better.

Please have another look.

mickenordin (approved):
Looks good!

PR #51, opened 2023-06-09
DOCKERHOST: Set default dns to 89.32.32.32

mickenordin:
This patch changes the default dns for docker that is set in
    /etc/default/docker from the host ip address to 89.32.32.32 which is a
    know working resolver operated by Sunet. This means that dns inside of
    docker works out of the box on debian.


theseal (approved):
Seems like a sane default 👍

theseal:
Approval withdrawn due to discussion on Slack regarding if this also affects unbound which might use `docker_dns`.

eest:
I am unsure how this interacts with machines setting `manage_dockerhost_unbound` to true, will need testing to figure out how other users of the `docker_dns` variable (like creating ufw rules) are affected by this and if it might break such setups.

mikaelfrykholm:
Jag sprang på detta och brände några timmar på att vi inte har en fungerande default dns för docker.

PR #52, opened 2023-06-19
Kano podman

mickenordin:
This patch adds support for using podman as a replacement for docker


manifests/packages/podman_compose.pp
```diff
@@ -0,0 +1,12 @@
+# Install podman-compose from pip
+class sunet::packages::podman_compose(
+) {
+  include sunet::packages::pip
+  include stdlib
+
+  exec { 'podman-compose':
+    command => 'pip install podman-compose',
```
theseal, line 8: Should probably be changed to `python3 -m pip install podman-compose`.

PR #54, opened 2023-06-29
Replace "iif" with "iifname"

eest:
Allows the ruleset to be loaded even if the interface mentioned in the rule does not exist yet. This is helpful when using interfaces like "wg0" which might be created after the ruleset is loaded during boot.


eest:
Have tested this on ubuntu 22.04 machines

PR #56, opened 2023-07-05
Cache assets to workaround Mastodon bug

theseal:


theseal:
Please don't squash 🙂

PR #59, opened 2023-07-10
Generic class to setup a KVM server with cloud images

theseal:
Bits and pieces from all kvm classes found in different ops repos.

This class does not create the required bridge interface. It unclear to me if it's wise to try to do that in here since it might interfere with other network configuration. E.g cnaas provides their own netplan configuration through plain files in `overlay` 
Any thoughts?

Don't think there is anything breaking in here so it's probably safe to just trail and error…


mickenordin:
Is this still relevant? Can @eest take a look maybe?

theseal:
Yes it's still relevant!

eest (approved):
Finally tested on norpan-kvm2, could see no effect on running instance. Also tested reboot and the guest comes up as expected.

LGTM 🎉

PR #62, opened 2023-07-10
Improve pinning of docker and more

theseal:
* Pin the whole repo so that we don't get packages not covered by `unattended-updates` floating around.
* `puppet-lint`
* Remove legacy transitions to `docker-ce` package
* Add tool to upgrade to the latest available docker version
* Default to install the latest docker version instead of an older configured one (not a breaking change since we previously required a docker version to run the class)


mickenordin (approved):
I think all of these changes look fine, they should not affect anything newer than Ubuntu 20.04 so nothing using the stable branch will be affected anyway

PR #67, opened 2023-07-13
Create authorized keys

mickenordin:
If the hash ssh_authorized_keys is defined, we should set up the corresponding resource. It seems now, everyone is doing this in cosmos-site.pp, which does not seem right.


theseal:
I'm all for a migration from `cosmos-sites.pp` but shouldn't it be better to embrace `sunet::ssh_keys` instead or is that just to complicated?

PR #68, opened 2023-07-13
Remove deprecated function calls to hiera and friends

mickenordin:
Switch hiera('blablabla') -> lookup('blablabla', undef, undef, undef)
Switch hiera('blablabla', 'NOT_SET_IN_HIERA') -> safe_hiera('blablabla')
Switch hiera('blablabla', 'blabla') -> lookup('blablabla', undef, undef, 'blabla')
Switch hiera_array('blablabla', ['blabla']) -> lookup('blablabla', undef, undef, ['blabla'])
Switch safe_hiera('blablabla', 'NOT_SET_IN_HIERA') -> safe_hiera('blablabla')
Switch safe_hiera('blablabla', 'blabla') -> lookup('blablabla', undef, undef, 'blabla')

Note: If types are known at compiletime we also add that like:

String $str = lookup('str', String, undef, '')

Also switch the safe_hiera function over to lookup function in ruby code

More info on this here: https://www.puppet.com/docs/puppet/5.5/hiera_automatic.html#looking-up-data-with-hiera

Note: do not squash, there are lint commits here as well

### DIFF
--- a/manifests/backuphost.pp
+++ b/manifests/backuphost.pp
@@ -3,7 +3,7 @@ class sunet::backuphost(
   String        $chroot,
   String        $mountpoint,
   String        $user   = 'backup',
-  Array[String] $allow_clients = hiera('backup_pfx'),
+  Array[String] $allow_clients = lookup('backup_pfx', Array[String], undef, ['130.242.125.68', '130.242.121.73']),
 ) {
   # parameters for sunet/backuphost/cron_free_diskspace.erb
   $free_diskspace_basedir = "${chroot}/incoming"
@@ -92,7 +92,7 @@ class sunet::backuphost(
     warn_criteria => ['exit_status=1'],
   }
 
-  $ssh_key_db = lookup('backup_ssh_keys')
+  $ssh_key_db = lookup('backup_ssh_keys', undef, undef, undef)
   if is_hash($ssh_key_db) {
     sunet::ssh_keys { 'backuphost':
       config   => { $user => keys($ssh_key_db) },
--- a/manifests/security/configure_sshd.pp
+++ b/manifests/security/configure_sshd.pp
@@ -3,7 +3,7 @@
 # @param port            The port to configure sshd to listen on. If undef, we leave it to sshd to decide.
 class sunet::security::configure_sshd (
   Boolean $configure_sftp = true,
-  Optional[Integer] $port = hiera('sunet_ssh_daemon_port', undef),
+  Optional[Integer] $port = lookup('sunet_ssh_daemon_port', Optional[integer], undef,  undef),
 ) {
   ensure_resource('package', 'openssh-server', {
       ensure => 'installed'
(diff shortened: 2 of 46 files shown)


manifests/backuphost.pp
```diff
@@ -3,7 +3,7 @@
   String        $chroot,
   String        $mountpoint,
   String        $user   = 'backup',
-  Array[String] $allow_clients = hiera('backup_pfx'),
+  Array[String] $allow_clients = lookup('backup_pfx', Array[String], undef, ['130.242.125.68', '130.242.121.73']),
```
mickenordin, line 6: I don't know if this is controversial or not, I don't want to lock out anyone from ssh, so set hoppjerka and joppherka here as default

mickenordin (commented):
Should be safe, one possible controversial change is commented on below

manifests/security/configure_sshd.pp
```diff
@@ -3,7 +3,7 @@
 # @param port            The port to configure sshd to listen on. If undef, we leave it to sshd to decide.
 class sunet::security::configure_sshd (
   Boolean $configure_sftp = true,
-  Optional[Integer] $port = hiera('sunet_ssh_daemon_port', undef),
+  Optional[Integer] $port = lookup('sunet_ssh_daemon_port', Optional[integer], undef,  undef),
```
theseal, line 6: Is this a valid syntax?

manifests/snippets/secret_file.pp
```diff
@@ -13,7 +13,7 @@
     default  => $path
   }
   $safe_key = regsubst($hiera_key, '[^0-9A-Za-z_]', '_', 'G')
-  $data = hiera($hiera_key, hiera($safe_key))
+  $data = lookup($hiera_key, undef, undef, lookup($safe_key, undef, undef, undef))
```
theseal, line 16: Inception?

theseal (approved):
Please have a look at my comments and make sure that syntax is correct.
Besides that I see no issue merging. Would have done it my self if it wasn't my last Friday before vacation. Merge with care and caution 🙂

manifests/snippets/secret_file.pp
mickenordin: You think it should be brokendown into two lines instead?

theseal: Two lines or maybe two variables but that might be what you ment? Easier to read and understand.

manifests/security/configure_sshd.pp
mickenordin: Yes. As far as I can tell, and as my testing can conclude.

https://www.puppet.com/docs/puppet/5.5/hiera_automatic.html#puppet_lookup

 lookup( <NAME>, [<VALUE TYPE>], [<MERGE BEHAVIOR>], [<DEFAULT VALUE>] ) 
<VALUE TYPE> (data Type) - A data type that must match the retrieved value; if not, the lookup (and catalog compilation) will fail. Defaults to Data which accepts any normal value.

PR #70, opened 2023-08-09
Merge classes from swamid-ops and fidus-ops

theseal:
New fancy classes to run a parts of a federation.


theseal:
Since all classes are new to puppet-sunet and I'm the only one intressted in them I will merge this PR right away.

PR #72, opened 2023-08-09
Add missing template

theseal:
All servers that where running the class had access to the swamid namespace and therefor worked just fine.


theseal:
LGTM

PR #73, opened 2023-08-10
Parent directory needs to be created on fresh install

mikaelfrykholm:
Tested on internal-sto1-test-forgejo-2.sunet.se/


theseal (approved):

theseal:
LGTM

PR #74, opened 2023-08-14
"Cherry-pick" `sunet::baas2` to legacy sunet branch

theseal:


theseal:
See #75 instead. Better to use a compatible syntax for all versions of baas2.

PR #76, opened 2023-08-29
Convert from pre puppet 4 legacy function api

mickenordin:
All these functions use a legacy format that is called "legacy" in puppet 5.5[0].

I have used the format described in the refactor guide to fix all of them[1].

0. https://www.puppet.com/docs/puppet/5.5/functions_legacy.html
1. https://www.puppet.com/docs/puppet/7/functions_refactor_legacy


theseal:
Tried it on a Debian 11 running Dehydrated and seems to run just fine. Your call for Ubuntu @eest

theseal (approved):

eest (approved):

PR #77, opened 2023-08-30
Move requirements to metadata file

mickenordin:
In puppet >= 7 we get an error when requiring classes across modules:

```
Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Could not find class ::concat for internal-sto3-test-imap-1.mail.sunet.se (file: /etc/puppet/cosmos-modules/sunet/manifests/ssh_keyscan/host.pp, line: 6, column: 3) (file: /etc/puppet/cosmos-modules/sunet/manifests/dehydrated.pp, line: 285) on node internal-sto3-test-imap-1.mail.sunet.se
```

This patch adds information that there is a dependency on the   puppetlabs-concat and puppetlabs-stdlib modules to metadata.json, and removes the require clauses from the manifests (that were allways unneccessary). The classes and other resources are still available as puppet will resolve them without explicitly requiring them in the manifest. This has been tested on puppet 7.23.0 and puppet 5.5.22.

### DIFF
--- a/manifests/exabgp/monitor.pp
+++ b/manifests/exabgp/monitor.pp
@@ -17,7 +17,6 @@ define sunet::exabgp::monitor::url(
   Integer          $prio  = 10,
   String           $path  = '/etc/bgp/monitor.d',
 ) {
-   require stdlib
    $check_url = $url ? {
       undef   => $name,
       default => $url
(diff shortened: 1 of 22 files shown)


theseal:
Seems to work in Debian 11.

theseal (approved):

theseal:
Approved by should (as #76 ) be tested in an modern Ubuntu as-well which I don't have access to.

eest (approved):
Tested on a dehydrated-running ubuntu 22.04
LGTM

PR #79, opened 2023-10-06
Remove deprecated includes for puppet 7

mikaelfrykholm:


mikaelfrykholm:
Tested only on debian12

PR #81, opened 2023-10-19
Make manage_ssh_key generate a valid key instead of creating a 0 byte…

mikaelfrykholm:
Tested on debian bullseye.


theseal:
Please run `puppet-lint` and push again.

theseal:
LGTM. Vi tar resten av lintningen en annan gång.

PR #83, opened 2023-10-24
Some fixes for puppet 7.x compat

mickenordin:
* Remove include statement from exabg.pp
* Split out defines from loadbalancer.pp into separate files with namespaces
* Split out dehydrated.pp into separate files
* Split out ssh_keyscan.pp into separate files


theseal (approved):
LGTM

PR #91, opened 2023-11-07
BankIdP + some docker stuff

theseal:


manifests/docker_compose.pp
mickenordin, line 1: This one I am not sure about... I would suggest something like:

```
  $nftenabled_and_interface =  $::facts['sunet_nftables_enabled'] == 'yes' and has_key($::facts['networking']['interfaces'], 'to_docker')
  $advanced_network_or_nftdisabled = $::facts['dockerhost_advanced_network'] == 'yes' or $::facts['sunet_nftables_enabled'] == 'no'
  if ( $nftenabled_and_interface or  $advanced_network_or_nftdisabled ) {
    $_install_service = true
  } else {
    $_install_service = false
    notice("sunet::docker_compose: Not installing ${service_name}, interface to_docker missing")
  }

  if $_install_service {
```

mickenordin (changes requested):
I revied the stuff about advanced network and didn't check the bankidp stuff :)

manifests/docker_compose.pp
mikaelfrykholm: Yes, I struggled with that logic as well. I failed when I tried re-setting variables.

mikaelfrykholm: your suggestion worked out of the box

mickenordin (approved):
Looks good!

manifests/rediscluster.pp
```diff
@@ -29,9 +32,11 @@
       ensure  => present,
       content => template('sunet/rediscluster/server.conf.erb'),
     }
-    -> sunet::misc::ufw_allow { "redis_port_${i}":
```
mickenordin, line 32: Should this be wrapped in:

```
if $::facts['sunet_nftables_enabled'] == 'yes'  or $::facts['dockerhost_advanced_network'] == 'yes' {
....
}
```

Leaving the old ufw rules inplace if those conditions are not met?

mickenordin (changes requested):

manifests/rediscluster.pp
mikaelfrykholm: Good catch, fixed!

mickenordin (approved):
lgtm

PR #93, opened 2023-11-14
Improved handling of NRPE and checks

theseal:


files/nagios/check_apt-wrapper
mikaelfrykholm, line 1: Why is this apt preferences stuff needed?

mikaelfrykholm: Looks good otherwise.

mikaelfrykholm (approved):

files/nagios/check_apt-wrapper
theseal: Without the wrapper and it's configuration the `check_apt` check will do false positives when an update is available before unattended updates has been run.

PR #97, opened 2023-11-20
Fix nrpe check process

theseal:
If `sunet::nagios::nrpe_command` was called multiple times `puppet` would crash since the file statement was run more then once.


mickenordin (approved):

mickenordin:
LGTM

PR #102, opened 2023-11-24
Fix gpg import to fogejo.

maandersson:
We have to use gpg batch mode to import a private key non interactive.


mickenordin (approved):
lgtm

PR #103, opened 2023-11-24
Forgeobackup: Fix variable interpolation in script template.

maandersson:
Opening tag for variable interpolation in template for backup script was missing a "=".


mickenordin (approved):
Nice catch!

PR #106, opened 2023-11-30
IBMs deb package is not compatible with Debian

theseal:
IBM have in their install scripts a special handle of Ubuntu where they link
all libs to /usr/lib which makes the dynamic linker find their libraries.

On Debian the dynamic linker can't find their libraries:
/usr/bin/dsmc: error while loading shared libraries: libgsk8ssl_64.so: cannot
open shared object file: No such file or directory

By adding this configuration to the linker it will extend it's search paths to
include where IBMs package install their libs.


SpaceFarmer (approved):
This looks good to me.

PR #107, opened 2023-11-30
Add ability to ignore non-critical errors

SpaceFarmer:
Add the ability to ignore errors that we consider to be non critical so we can avoid unnecessary alarms in scriptherder.


eest (approved):
LGTM

PR #114, opened 2023-12-04
Notify/subscribe are executed to late

theseal:
We need to update the library cache as soon as possible. It seems like the notify/subscribe commands are run to late in the process so the boostrap script failed due to lack of libraries.

By running the needed commands in a "regular" exec the order seems to be garanteed.


SpaceFarmer (approved):
Looks good to me.

eest (approved):
LGTM

PR #116, opened 2023-12-15
Fix as per https://github.com/swedenconnect/bankid-saml-idp/issues/233

mikaelfrykholm:
Small fix in upstream examples what we need to update our templates for.


theseal (approved):
LGTM

PR #118, opened 2024-01-09
Allow to override thruk shibboleth config.

mikaelfrykholm:
Good sirs, I have created a way to override the shib config by providing the path to file to be mounted in the container.


theseal (approved):
LGTM

PR #119, opened 2024-01-17
We can't operate on None objects

theseal:
If `sunet::dockerhost` is configured in `cosmos-rules.yaml` without any options the class becomes a None object.


mikaelfrykholm (approved):
lgtm

PR #123, opened 2024-01-31
Modernize configuration template

theseal:
Big win with this version is that unattended-updates now (again) include the Debian Security repo which was renamed in 2019:
https://lists.debian.org/debian-devel-announce/2019/07/msg00004.html


mikaelfrykholm (approved):
lgtm

PR #124, opened 2024-02-02
Safe defaults

theseal:


mickenordin (approved):
As discussed I think this is a very good idea :). Implementation looks fine too.

PR #125, opened 2024-02-06
Automatically migrate docker_run to docker-compose

theseal:
When `dockerhost2` is configured as the docker class of choice.

This PR contains a demo: https://play.sunet.se/media/Automatic+migration+from+docker_run+to+to+docker-compose+with+dockerhost2/0_dpudtaz8


mickenordin (approved):
Very slick!

PR #131, opened 2024-02-16
Mifr redis rerelayout

mikaelfrykholm:
Revert to the old way of clustering as the failover did not work as intended.


theseal (approved):
LGTM

PR #132, opened 2024-02-19
Prevent reboot if master for multiple shards

theseal:


manifests/rediscluster.pp
mikaelfrykholm, line 1: We probably should hide this behind a featureflag. It assumes that we run TLS between nodes and that the certificates are at a specific location.

mikaelfrykholm (changes requested):

mikaelfrykholm (approved):

PR #136, opened 2024-02-26
Dockerize the mariadb class

theseal:


mickenordin (approved):
This looks like a fine manifest for all your mariadb needs.

PR #137, opened 2024-02-27
Create a default DB using variables

theseal:
Make is possible to use the variables the image honors to create a default database and user.


mickenordin (approved):
lgtm

PR #140, opened 2024-03-13
Do not print lic key to reduce risk of leaking it

SpaceFarmer:
Alter print statements to not include license key and contract id. This is done to reduce the risk of it ending up in for example syslog.


eest (approved):
LGTM

PR #141, opened 2024-03-14
A mixed bag

theseal:
* A new class to handle acme-d througt certbot. Will replace https://wiki.sunet.se/display/sunetops/How+to+create+a+cert+with+acme-d
* A new class (WIP) to setup get eduroam
* A few other tweaks…


files/certbot/certbot-acmed-renew-post-hook-wrapper
eest, line 1: Would we want to store this in a separate dir, something like /etc/sunet-certbot-acmed instead? Just wondering if it is OK to do your own thing under /etc/letsencrypt or not.

theseal: Problem solved by reading the docs and getting help from mifr. We new uses certbots own hook dir(s).

eest (approved):
LGTM

mikaelfrykholm (approved):
LGTM

PR #143, opened 2024-03-20
Allow multiple certs

mickenordin:
With this patch set you can add multiple certificates to tls_certificates_bundle in this format:

```
tls_certificate_bundle: "/etc/ssl/private/kube.drive.test.sunet.dev_haproxy.crt crt  /etc/ssl/private/kube.drive.sunet.se_haproxy.crt"
```

That is, separated by the string ' crt '

The change is backwards compatible with the old solution and has been tested on drive frontends. Because of the usage of ' crt ' as seprator, {{ tls_certificate_bundle }} can still be used in your templates as before.

### DIFF
--- a/manifests/frontend/load_balancer/website2.pp
+++ b/manifests/frontend/load_balancer/website2.pp
@@ -70,7 +70,7 @@ define sunet::frontend::load_balancer::website2(
   $numcerts = length($temp_certs)
   if $numcerts > 1 {
     $certnum = 0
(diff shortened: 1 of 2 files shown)


johanlundberg:
LGTM

johanlundberg (approved):
LGTM

theseal (approved):
I don't know enough about HAproxy be puppet and docker config seems sane.

PR #145, opened 2024-03-27
Use local freedesktop_os_release() on python < 3.10

SpaceFarmer:
Turns out we need the local function on python 3.8 as well. That is the default for ubuntu-20.
This function is only available from python 3.10:
https://docs.python.org/3/library/platform.html#platform.freedesktop_os_release


eest (approved):
LGTM

PR #146, opened 2024-04-02
Starship

theseal:
Attempt to replace eduid's (sometimes) broken prompt


mikaelfrykholm (approved):
LGTM, sha256 sum verified with upstream.

johanlundberg (approved):
As discussed with @theseal on Slack. I would like, for once, to have this change i a single commit to make it easier to cherrypick.

PR #148, opened 2024-04-11
Due to Jsch bugs when only allowing ssh-ed25519 we use ecdsa for jenk…

mikaelfrykholm:
…ins->pypi.
https://github.com/jenkinsci/publish-over-ssh-plugin/issues/341


theseal (approved):
LGTM

PR #153, opened 2024-04-26
Adjust test of dmi=>product=>name to match that of DCO.

maandersson:
Adjust test to match value for DCO so that nodes there get puppet class sunet::iaas::server.


mickenordin (approved):
Looks good!

PR #157, opened 2024-05-14
Service override to create missing dir caused by container entrypoint override.

maandersson:
Creates missing directory /tmp/gitea inside the forgeo container at start/restart of the systemd service.
An service override file creates ExecStartPost actions to do this.
This directory is missing because we override the container entrypoint.
This is a workaround until we have a better solution in place.
We should not merge to stable until we have a service window because it will cause a restart of the service.


mikaelfrykholm (approved):
LGTM

PR #159, opened 2024-05-23
updated scriptherder check for sunet infra cert

mariahaider:
Changes:

- Make the monitoring optional for services that do not bother about the validity of the certs
- Make the scriptherder check criticial if the cert is going to expire within two weeks or it is completely expired
- Warning for scriptherder is removed as the scriptherder check never becomes critical
- Adjust the script according to the scriptherder check

### DIFF
--- a/manifests/ici_ca.pp
+++ b/manifests/ici_ca.pp
@@ -42,10 +42,13 @@ define sunet::ici_ca::autosign(
 }
 
 # fetch certificate from ici-ca automatically and run a scriptherder job to see if it is valid
-define sunet::ici_ca::rp()
-{
+define sunet::ici_ca::rp(
+  Boolean $monitor_infra_cert = true,
+) {
+
   $host = $::fqdn
   $ca = $name
+
   file { '/usr/bin/dl_ici_cert':
     content => template('sunet/ici_ca/dl_ici_cert.erb'),
     mode    => '0755'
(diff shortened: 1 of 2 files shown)


theseal:
> Changes:
> 
> * Make the monitoring optional for services that do not bother about the validity of the certs

Great!

> * Make the scriptherder check criticial if the cert is going to expire within two weeks or it is completely expired
Why remove the warning mechanism? Better to be able to set different configurations in puppet

> * Warning for scriptherder is removed as it doesn't seem to work properly
> * Adjust the script according to the scriptherder check

What do you mean? I just tried with an example check and it works just fine. Could it be that scriptherder was told to warn if exit status was `2` and the check exited `1` since it was just a warning?

mariahaider:
> > Changes:
> > 
> > * Make the monitoring optional for services that do not bother about the validity of the certs
> 
> Great!
> 
> > * Make the scriptherder check criticial if the cert is going to expire within two weeks or it is completely expired
> 
> Why remove the warning mechanism? Better to be able to set different configurations in puppet
> 
> > * Warning for scriptherder is removed as it doesn't seem to work properly
> > * Adjust the script according to the scriptherder check
> 
> What do you mean? I just tried with an example check and it works just fine. Could it be that scriptherder was told to warn if exit status was `2` and the check exited `1` since it was just a warning?

It does warn but it never becomes critical. I tried in different combinations, I checked with exit status 2 and 0.

mariahaider:
yes, it would be nice. Then I can update it later. But I felt that it was better that check would become critical after a point to urge people to change it. So I chose to make it criticial. What is your suggestion?

theseal (approved):
Yes, please revert the check to support warnings if scriptherder getting support for signaling critical in combination with `max_age`.

PR #160, opened 2024-05-24
Mifr monitor changes

mikaelfrykholm:
First draft of monitor changes for accepting OpenTelemetry data.


templates/naemon_monitor/mimir-server.yaml
```diff
@@ -0,0 +1,47 @@
+# Do not use this configuration in production.
```
eest, line 1: Should the comment that this should not be used in production still be here? Do we need to tweak something in order to make it "production ready"?

eest (commented):
See some comments inline

templates/naemon_monitor/mimir-server.yaml
mikaelfrykholm: Copy paste error. That was from an example storing persistent data i /tmp

mikaelfrykholm:
I fixed owner attributes of files and also included some rudimentary dashboards.

theseal (approved):
LGTM

PR #161, opened 2024-05-24
Some NRPE checks will get fishy results when using a PrivateTmp.

theseal:
E.g check_apt: https://askubuntu.com/questions/1415415/check-apt-issue-with-nagios


theseal:
Not tested on a new machine.

mariahaider:
Looks good to me but should it apply only to servers running 22.04?

mickenordin (approved):
Looks good!

theseal:
> Looks good to me but should it apply only to servers running 22.04?

It's needed for Debian aswell.

PR #162, opened 2024-06-19
ubuntu24 fixes

SpaceFarmer:
Ubuntu24 runs a modern puppet version (8.4.0) by default and some puppet syntax used in modules/manifests has been deprecated and no longer works.

Some useful info on the subject:
https://www.puppet.com/docs/pe/2023.6/osp/upgrading-from-puppet7-to-puppet8#upgrading-from-puppet7-to-puppet8
https://github.com/puppetlabs/puppet-lint/blob/main/lib%2Fpuppet-lint%2Fplugins%2Flegacy_facts%2Flegacy_facts.rb


mickenordin (approved):
Looks good to me!

PR #164, opened 2024-06-28
ubuntu24 fixes

SpaceFarmer:
- Fixes for deprecated puppet syntax
- Fixes for libvirt 10 (default in ubuntu24)
- Add 'check_needrestart' that checks which daemons need to be restarted after library upgrades

### DIFF
--- a/manifests/nagios/nrpe_check_total_procs_lax.pp
+++ b/manifests/nagios/nrpe_check_total_procs_lax.pp
@@ -3,7 +3,7 @@ define sunet::nagios::nrpe_check_total_procs_lax (
   Integer $critical = 200,
   Integer $warning = 150,
 ) {
-  if 'cosmos' in $facts and ('frontend_server' in $facts['cosmos']['host_roles']) {
+  if $facts =~ Hash and 'cosmos' in $facts and ('frontend_server' in $facts['cosmos']['host_roles']) {
     # There are more processes than normal on frontend hosts
     $_procw = $warning + 350
     $_procc = $critical + 350
--- a/manifests/nagios/nrpe_check_total_procs_lax.pp
+++ b/manifests/nagios/nrpe_check_total_procs_lax.pp
@@ -3,7 +3,7 @@ define sunet::nagios::nrpe_check_total_procs_lax (
   Integer $critical = 200,
   Integer $warning = 150,
 ) {
-  if is_hash($facts) and has_key($facts, 'cosmos') and ('frontend_server' in $facts['cosmos']['host_roles']) {
+  if 'cosmos' in $facts and ('frontend_server' in $facts['cosmos']['host_roles']) {
     # There are more processes than normal on frontend hosts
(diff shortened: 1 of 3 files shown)


manifests/nagios/nrpe_check_total_procs_lax.pp
```diff
@@ -3,7 +3,7 @@
   Integer $critical = 200,
   Integer $warning = 150,
 ) {
-  if is_hash($facts) and has_key($facts, 'cosmos') and ('frontend_server' in $facts['cosmos']['host_roles']) {
+  if 'cosmos' in $facts and ('frontend_server' in $facts['cosmos']['host_roles']) {
```
eest, line 6: Here we are removing two operations but only replacing with one: has_key() seems reasonable to replace with "if in" syntax, but "is_hash()" is just removed. Is this on purpose? Can this lead to doing "if in" on some other variable type like a string?

templates/kvm/mk_cloud_image.erb
```diff
@@ -70,8 +72,10 @@ fi
 # solve bootstrapping of new kvm hosts where no vms exist already
 virsh pool-list | grep -qe "^\s*${pool_name}.*active" || {
     virsh pool-create-as --name "${pool_name}" --target "${images_dir}" --type dir --build
-    virsh pool-autostart "${pool_name}"
-    virsh pool-start "${pool_name}"
+    if dpkg --compare-versions ${VERSION_ID} lt 24.04 ; then
```
eest, line 75: This throws a shellcheck warning:
```
if dpkg --compare-versions ${VERSION_ID} ge 24.04 ; then
                           ^-----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

templates/kvm/mk_cloud_image.erb
```diff
@@ -105,6 +109,9 @@ fi
 if [ "x${rng}" != "x" ]; then
    install_options="${install_options} --rng='${rng}'"
 fi
+if dpkg --compare-versions ${VERSION_ID} ge 24.04 ; then
```
eest, line 112: Same shellcheck varning as above.

eest (commented):
The adding of a new class seems OK, some other comments inline.

templates/kvm/mk_cloud_image.erb
SpaceFarmer: Added double quotes

templates/kvm/mk_cloud_image.erb
SpaceFarmer: Added double quotes

manifests/nagios/nrpe_check_total_procs_lax.pp
SpaceFarmer: Added a check to make sure it is a hash like it was before, but with new syntax.

eest (approved):
LGTM

PR #165, opened 2024-06-28
Thorslund lb dockerhost2

tuxdepend:
Fixed manifests for new load balancers.


theseal:
Added Maria and kano as reviewer as they are more involved in the loadbalancers. One question from my perspektive. Why do you need network-mode `host` for the containers?

mariahaider:
@mickenordin Regarding multiple frontends, currently we have two frontend instances, one fully tested, the other is being set up.
If you are refering to the `network_mode: host` bit, in that case, it is only `frontend-api` and `frontend-telegraf` containers who will be using host network. They run as single containers. The frontand instances as usual will communicate through `br-<instance>` docker network bridge. We haven't changed the network setup for those.

mariahaider:
@mickenordin @theseal do you have any other comments?

theseal:
> @mickenordin Regarding multiple frontends, currently we have two frontend instances, one fully tested, the other is being set up. If you are refering to the `network_mode: host` bit, in that case, it is only `frontend-api` and `frontend-telegraf` containers who will be using host network. They run as single containers. The frontand instances as usual will communicate through `br-<instance>` docker network bridge. We haven't changed the network setup for those.

I think @mikaelfrykholm is working on a solution to fix the NAT issues you describing. I would recommend you to switch back to `regular` network_mode when that is in place.

> @mickenordin @theseal do you have any other comments?

Nope.

mariahaider:
@theseal will do, will be a bit tricky after it goes into production though but we will have redundance.

PR #166, opened 2024-06-29
Make sure we don't change behaviour for debian

SpaceFarmer:
Make if statements more precise so they do not affect _debian_ since that is not tested.

**if statement tests:**
```
$ cat test.sh
#!/bin/bash
#

. /etc/os-release

echo "ID: $ID"
echo "VERSION_ID: $VERSION_ID"

if [ "${ID}" = "debian" ] || ([ "${ID}" = "ubuntu" ] && dpkg --compare-versions "${VERSION_ID}" lt 24.04) ; then
  echo "ubuntu and less than 24.04 or debian"
else
  echo "no match"
fi
```
**ubuntu22:**
```
$ ./test.sh
ID: ubuntu
VERSION_ID: 22.04
ubuntu and less than 24.04 or debian
```
**ubuntu24:**
```
$ ./test.sh
ID: ubuntu
VERSION_ID: 24.04
no match
```
**debian12:**
```
$ ./test.sh
ID: debian
VERSION_ID: 12
ubuntu and less than 24.04 or debian
```

### DIFF
--- a/templates/kvm/mk_cloud_image.erb
+++ b/templates/kvm/mk_cloud_image.erb
@@ -72,7 +72,7 @@ fi
 # solve bootstrapping of new kvm hosts where no vms exist already
 virsh pool-list | grep -qe "^\s*${pool_name}.*active" || {
     virsh pool-create-as --name "${pool_name}" --target "${images_dir}" --type dir --build
-    if dpkg --compare-versions "${VERSION_ID}" lt 24.04 ; then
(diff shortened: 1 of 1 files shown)


theseal:
Don't think we have any KVMs on Debian but code looks good 👍

eest (commented):
LGTM, thanks for showing test cases!

PR #167, opened 2024-07-04
pahol-ici

SpaceFarmer:
- Fixes for ici CA to be able to run on ubuntu22
- Fix for ::dockerhost2 where the docker network was not created


theseal (approved):
Good enough workaround for this service which soon reaches EOL.

PR #168, opened 2024-07-12
Thorslund lb dockerhost2

tuxdepend:
We replaced the https://github.com/SUNET/puppet-sunet/blob/main/templates/dockerhost/check_docker_containers.erb with https://github.com/SUNET/puppet-sunet/blob/thorslund-lb_dockerhost2/templates/dockerhost/check_docker_containers_noversion.erb for hosts running dockerhost2 as 'version' is depracated in docker compose.


theseal (changes requested):
In order to not needing to maintain two diffrent version of the same script I would prefer if the check would handle/check if it is running under `dockerhost2` OR if the template would skip to include the version check if run under `dockerhost2`. The later is probably easier since there is a fact for `dockerhost2` present.

theseal (approved):
LGTM

PR #170, opened 2024-07-17
/etc/motd.tail not been in used since 2014

theseal:
Not sure if anyone really cares about this class since it has been broken since around 2014 🤷‍♂️

See Debians documentation: https://wiki.debian.org/motd


mikaelfrykholm (approved):
LGTM

mickenordin (approved):

PR #172, opened 2024-08-05
Mifr container registry fix

mikaelfrykholm:
Work around problem with pushing to registry gives "413 Request Entity Too Large". Not exhaustively tested to verify if we have a problem with chunked uploads.


eest (approved):
Maybe reference https://docs.gitea.com/administration/reverse-proxies#nginx as a source for the 512MB starting point? LGTM either way.

PR #173, opened 2024-08-07
Initial commit of sunet::chrony

eest:
This is expected to be a replacement for sunet::ntp on Ubuntu 23.04 and newer as well as Debian 12 and newer (via later modification of sunet::server).

This module uses a default set of time sources from netnod to adhere to MSBFS 2020:7 which states machines should use time traceable to the official Swedish time UTC(SP).

It also defaults to using NTS over NTP to make our servers more resilient against MITM attacks trying to feed us bad time.

The chrony.conf template is basically a copy of the one supplied by Ubuntu/Debian with a few changes:
* Remove the default pool directives used by the Ubuntu/Debian configs and handle our own time sources by creating a sunet.sources file in the sources.d dir. These files can be reread via `chronyc reload sources` instead of restarting chrony.
* Set "cmdport 0": this removes a listening port 323 for chronyc commands. This is one less port to wonder about in netstat output and chronyc still works because it uses a unix socket by default.
* Toggle the usage of DHCP-learned NTP sources with a "dhcp_time_sources" argument, false by default. Since we want to use NTS it seems bad to possibly mix in ordinary NTP sources if a DHCP server happens to advertise them on a given network.

Finally you can also act as NTP server by adding entries to the allow_list argument which will make chrony open listening ports. This server operation can be further modified to handle NTS operation by supplying a NTS cert and key. Handling of such certs (e.g. via certbot or similar) is currently out of scope, but can possible be combined with something like sunet::certbot::acmed for automatic managemnt in the future.

### DIFF
--- a/manifests/chrony.pp
+++ b/manifests/chrony.pp
@@ -1,0 +1,53 @@
+# Install and configure chrony service. Default setup is client-only using NTS
+# time sources. If you want to act as NTP server see $allow_list.
+# Using the netnod NTS servers by default keeps us in line with MSBFS 2020:7
+# with their time service being traceable to UTC(SP).
+class sunet::chrony(
+  Array[String] $ntp_servers = [],
+  Array[String] $nts_servers = [
+    'gbg1.nts.netnod.se',
+    'gbg2.nts.netnod.se',
+    'lul1.nts.netnod.se',
+    'lul2.nts.netnod.se',
+    'mmo1.nts.netnod.se',
+    'mmo2.nts.netnod.se',
+    'sth1.nts.netnod.se',
+    'sth2.nts.netnod.se',
+    'svl1.nts.netnod.se',
+    'svl2.nts.netnod.se',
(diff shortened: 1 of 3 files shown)


manifests/chrony.pp
```diff
@@ -0,0 +1,53 @@
...
+    ensure  => 'file',
+    mode    => '0644',
+    owner   => 'root',
+    group   => 'root',
+    content => template('sunet/chrony/chrony.conf.erb'),
+    notify  => Service['chrony'],
+  }
+
+  exec { 'chronyc reload sources':
+      refreshonly => true,
```
mikaelfrykholm, line 42: wrong indent, should be 4

mikaelfrykholm (approved):
LGTM

manifests/chrony.pp
eest: Fixed, thanks for spotting it :)

PR #174, opened 2024-08-09
Only install rng-tools on older kernels

eest:
When working on a ubuntu 24.04 machine I saw this message on each cosmos run:
```
Notice: /Stage[main]/Sunet::Tools/Package[rng-tools]/ensure: created
```
This is because rng-tools is now a virtual package installing rng-tools-debian instead. Looking into this it seems rng-tools is not even useful for modern kernels, from
https://wiki.archlinux.org/title/Rng-tools:
```
Note: rng-tools is not needed anymore since Kernel 5.6 because /dev/random does not block anymore.
```
... so instead of pointing to rng-tools-debian or rng-tools5 on modern ubuntu machines just stop installing it for all machines with a kernel of 5.6 or later.

### DIFF
--- a/manifests/tools.pp
+++ b/manifests/tools.pp
@@ -13,11 +13,25 @@ class sunet::tools {
   $extra_tools = $facts['os']['name'] ? {
     'Ubuntu' => ['update-manager-core',
                  'unattended-upgrades',
-                 'rng-tools',
                  ],
-    'Debian' => ['rng-tools5',
-                ],
     default => []
   }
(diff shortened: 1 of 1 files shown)


eest:
Have not had a chance to test this on a debian machine yet, will try to track one down next week.

mickenordin:
This looks good to me, if you merge it to testing-branch and bump there it will end up on some debian machines.

eest:
Found a debian machine to test on, works as expected. This will of course only take effect for new machines, it will not clean up the package for already existing ones.

eest:
@mickenordin thanks for the review!

PR #175, opened 2024-08-12
Deprecate all uses of our old randomness services.

mikaelfrykholm:


btmattsson:
Är vi säkra på att gamla Ubuntu burkar inte behöver pollinate ?

mikaelfrykholm:
Kanske @mariahaider kan gräva nån nån riktigt gammal ubuntu vi kan testa på? Alla som kör Debian är det redan testat på. ubuntu >= 22.04 också via swamid.

eest:
Since the cron job is left behind, wont this trigger alerts when the binary is no longer present?

mikaelfrykholm:
Good catch. Fixed.

eest:
Removing the entry will stop it from being created, but it will not remove it from host that already have it. There will instead be an orphaned cron entry still trying to run. I think to remove it you will need to set it as ensure => absent, see https://www.puppet.com/docs/puppet/5.5/types/cron.html#cron-attribute-ensure

mikaelfrykholm:
Fixed!

theseal (approved):
LGTM

mariahaider:
We still have some some servers running 18, for example - ipam.sunet.se ( no plan to upgrade, needs to be replaced), ni.sunet.se and other NI servers ( will be upgraded)

mariahaider:
But they all run old 'sunet' branch of course.

PR #181, opened 2024-08-23
Transition MDQ publisher to docker compose

theseal:
[SWAM-59](https://jira.sunet.se/browse/SWAM-59)

There will be a 10 - 30 sec downtime of the publisher during the change so the feature flag should be turned on with caution.


mikaelfrykholm (approved):
LGTM

btmattsson (approved):
Ser OK ut.

PR #183, opened 2024-08-28
merge pahol-ubuntu24 fixes

SpaceFarmer:
Fix legacy facts and syntax that is incompatible with puppet8.x *(default in ubuntu24).


theseal (approved):
LGTM

PR #184, opened 2024-08-29
open just allowed ports

mariahaider:
If the host runs `dockerhost2` and uses `sunet::nftables::docker_expose` to add rules, then `sunet::nftables::allow` is used. I added the logic that it will check if `$allow_clients` carry values. If it does, then it will restrict the opening of the port only to those clients. Additionally I added the paramete `proto` which was missing earlier.


theseal (approved):
LGTM

PR #188, opened 2024-09-05
microk8s fixes

mikaelfrykholm:
Problems fixed:
* able to use separate worker nodes
* Work with multiple ip-addresses per node (needed for ipv6)

calico still creates legacy rules, but at least this gets us a working cluster.


mickenordin (approved):
I haven't tested it, but looks good

PR #189, opened 2024-09-06
Small fixes for LGTM stack.

mikaelfrykholm:
Thank you @masv3971 for finding bugs!
* Wrong permission on loki data dir.
* Listen on all interfaces to make docker sending data to localhost easier (We don't open external firewall)
* Tempo does not need to use a certificate locally
* Forgot to add a uid on datasource so dashboards do not work out of the box.


theseal (approved):
Don't know much about this stack but the changes LGTM

PR #191, opened 2024-09-20
Initial attempt of tab completion for scriptherder

theseal:


mickenordin:
I don't have much experience with writing bash completion, but I am wondering if we would benefit from using:

https://pypi.org/project/argcomplete/

to automatically generate completions?

theseal:
Oh. Never heard of about that project! 

Writing this PR in shell was a mess so I'm all for trying alternatives for easier maintainace. Let me give it a shot.

theseal:
As `scriptherder` is structured today we can't use `argcomplete` to get tab completion. Might be doable with a bigger rewrite which I can't do right now. 

Suggesting that we merge this PR and therefor get a basic completion.

mickenordin (approved):

PR #193, opened 2024-10-01
Monitor expiration of GPG keys

theseal:
Both for cosmos and metadata repo's.

### DIFF
--- a/files/nagios/check_gpg_keys.sh
+++ b/files/nagios/check_gpg_keys.sh
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
-set -o pipefail
+set -uo pipefail
 
-if [ -z "$1" ]; then
-	echo "UNKOWN: A directory is required as \$1"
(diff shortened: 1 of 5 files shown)


mikaelfrykholm (approved):
LGTM. 
Some small nitpicks.
* Maybe set -euo also 
* My backup key is not parsed. It seems to assume only one pub key in each file.

theseal:
Will look in to the other `set`s

Why do you have multiple keys? Tested this in multiple repos and the few files with multiple public keys where a misstake.

mikaelfrykholm:
My backup key uses the same email address as my primary one and when I exported that one it tagged along. I kind of like only having one file for primary and backup key.

mickenordin (approved):
I think this makes a lot of sense. I also think one key per file, makes sense too, since it makes it easy to spot mistakes, where someone added more than the intended key. It is not that hard to maintain one file per key you need.

PR #195, opened 2024-10-03
stop docker adding ipv6 rules

mariahaider:
docker-ce debian package version 5:27 adds ipv6 rules in nftables if the server or docker service is restared. We might as well stop it adding ipv6 rules along with ipv4 which is already in the configuration.


theseal:
LGTM

theseal (approved):

mikaelfrykholm (approved):
LGTM
I must say docker do not have the best "Do not break things in production" procedure.

PR #198, opened 2024-10-10
Smol fix to create a keyrings folder on older systems. 

mikaelfrykholm:
Tested on debian11, but should apply to older ubuntu as well.


mikaelfrykholm:
The failing tests is a lie! I fixed them in testing and main. There was a problem with our ruby installation.

theseal (approved):
LGTM

PR #201, opened 2024-10-11
Clean up

theseal:
Clean up old attempt of a better/cleaner exit for Naemon. This is handled by the container nowadays.


theseal:
Rediscovered an old issue (https://github.com/naemon/naemon-livestatus/issues/117) which I know a read multiple times before…
So the reload is a little bit of a workaround but is still better then bring down and up the whole stack.

mikaelfrykholm (approved):
LGTM, changed merge to testing branch as per new workflow.

mickenordin (approved):

PR #202, opened 2024-10-11
Randomize requests to acme-c.sunet.se

mariahaider:
Randomize the minute for sending request to acme-c.sunet.se by servers which run the puppet class with `single_domain: false`

### DIFF
--- a/manifests/dehydrated/client_define.pp
+++ b/manifests/dehydrated/client_define.pp
@@ -56,7 +56,7 @@ define sunet::dehydrated::client_define(
     ensure_resource('sunet::scriptherder::cronjob',  "dehydrated_fetch_${server}", {
       cmd    => "sh -c 'ssh -Ti \$HOME/.ssh/id_${_ssh_id} root@${server} | /bin/tar xvf - -C /etc/dehydrated/certs && /usr/bin/le-ssl-compat.sh'",
       user   => $user,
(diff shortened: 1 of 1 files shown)


eest:
I am not sure this works as expected. If `fqdn_rand()` returns "1" this will lead to the minute entry `*/1` which means "run every minute". I am not sure how the rest of the cron entry looks but if all other fields are `*`, would it be enough to run once an hour instead of every 20 minutes? If that is the case then we can probably replace `"*/${minute_random}"` with `"${minute_random}"` which then will mean "run at this minute (every hour)", and then no matter what rand returns the frequency will be the same, just distributed over the full hour.

eest:
Looking at https://www.puppet.com/docs/puppet/5.5/function.html#fqdn_rand it looks like it is exclusive:

> Generates a random Integer number greater than or equal to 0 and less than MAX

... so using `60` is probably correct.

mariahaider:
> ... and if we where to change it to `"${minute_random}"`, keep in mind that the valid range of the minute interval is `0-59`, so in that case we might need to change the value in `fqdn_rand(60)` (I dont recall if the rand range is inclusive (0-60) or exclusive (0-59) of the given number).

Good point, fixed it, running once an hour should be fine.

eest (approved):
LGTM

PR #206, opened 2024-10-21
Replace ntp with chrony on our servers

eest:
This affects Ubuntu 23.04 and up as well as Debian 12 and up.

It is unfortunate that we need to modify sunet::ntp as well, but the problem is that ops-repos sometimes pull in sunet::ntp by hand (sometimes to set specific configuration in cosmos-rules etc) so in those cases it is hard to do a clean cutover between them.

For that reason sunet::ntp now does nothing on hosts matching the same criteria as the ones where sunet::server applies sunet::chrony. This way both can be applied to a host at the same time.

### DIFF
--- a/manifests/ntp.pp
+++ b/manifests/ntp.pp
@@ -12,9 +12,8 @@ class sunet::ntp(
   # Only do anything on older Ubuntu and Debian. For later versions
   # sunet::server will use sunet::chrony instead. This if-statement is needed
   # because not every ops-repo uses sunet::server to select what ntp class to
-  # use. If changing this if-statement you might need to update it in
-  # sunet::server as well.
-  if ($distro == 'Ubuntu' and versioncmp($release, '22.04') <= 0) or ($distro == 'Debian' and versioncmp($release, '11') <= 0) {
+  # use.
+  if $::facts['sunet_chrony_enabled'] != 'yes' {
     # Help Puppet understand to use systemd for Ubuntu 16.04 hosts
     if $distro == 'Ubuntu' and versioncmp($release, '15.04') >= 0 {
       Service {
--- a/manifests/ntp.pp
+++ b/manifests/ntp.pp
@@ -7,7 +7,7 @@ class sunet::ntp(
 
   # Get facts for distro/release
   $distro = $facts['os']['distro']['id']
-  $release = $facts['os']['distro']['release']['full']
+  $release = $facts['os']['distro']['release']['major']
 
   # Only do anything on older Ubuntu and Debian. For later versions
   # sunet::server will use sunet::chrony instead. This if-statement is needed
--- a/manifests/ntp.pp
+++ b/manifests/ntp.pp
@@ -9,62 +9,69 @@ class sunet::ntp(
   $distro = $facts['os']['distro']['id']
   $release = $facts['os']['distro']['release']['full']
 
-  # Help Puppet understand to use systemd for Ubuntu 16.04 hosts
-  if $distro == 'Ubuntu' and versioncmp($release, '15.04') >= 0 {
-    Service {
-      provider => 'systemd',
+  # Only do anything on older Ubuntu and Debian. For later versions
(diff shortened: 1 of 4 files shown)


mikaelfrykholm:
Why not debian 11?

eest:
@mikaelfrykholm these are the versions where the `ntp` package went from installing classic ntpd to becoming a transitional package pulling in `ntpsec` so that basically decided the scope (because this is what made `sunet::ntp` break for ubuntu machines). I guess I do not really see a problem with lowering the version for debian if wanted, but it has not been tested. If the general consensus is that it should be enabled on debian 11 it would be nice to test it somewhere. The source branch is tagged with `patlu-use-chrony-2*` so i guess we could test via that branch in that case?

eest:
... maybe we can focus on getting this in as-is first and then we can look at enabling it for older stuff later if needed?

mikaelfrykholm:
We have some debian 11 i platform-ops which run the testing branch. I think that is good enough testing.

mikaelfrykholm (approved):
LGTM

manifests/ntp.pp
```diff
@@ -7,64 +7,71 @@
...
-  # Help Puppet understand to use systemd for Ubuntu 16.04 hosts
-  if $distro == 'Ubuntu' and versioncmp($release, '15.04') >= 0 {
-    Service {
-      provider => 'systemd',
+  # Only do anything on older Ubuntu and Debian. For later versions
+  # sunet::server will use sunet::chrony instead. This if-statement is needed
+  # because not every ops-repo uses sunet::server to select what ntp class to
+  # use. If changing this if-statement you might need to update it in
+  # sunet::server as well.
+  if ($distro == 'Ubuntu' and versioncmp($release, '22.04') <= 0) or ($distro == 'Debian' and versioncmp($release, '11') <= 0) {
```
mickenordin, line 17: Can we make a fact out of this, in sunet::ntp and use that in sunet::server, so if this fact is set, we do one thing, and another if the fact is unset? Meaning we can detect if the sunet::ntp package has been pulled in from the side?

mickenordin (changes requested):
If we make a fact in stead of using the same if statement I think we can avoid some future headaches.

manifests/ntp.pp
mickenordin: I have been discussing this with patlu and I think maybe syncing the if clause is a better way to go, facts can be finicky to get right too.

mickenordin (approved):

manifests/ntp.pp
eest: I guess we could, i agree keeping things in sync across two files by hand is not a great approach, however i don't feel it is necessarily worth the work of creating a custom fact script at this point as I dont think we will go back to ntpd and sunet::ntp will fade into the background the more we use never machines.

theseal (approved):
LGTM

Tested this on a Debian 12 which worked fine. One notice in that graph from `check_nrpe` against `ntp.se` that the offset is more fluctuating than with `ntpd`. Still within acceptable margin but an unexpected outcome.

Screenshot 2024-10-22 at 09 39 05

mickenordin (approved):
Very nice!

manifests/ntp.pp
eest: ... on the other hand, "lets do it right", now fact has been commited. Thanks for the pointer!

eest:
> LGTM
> 
> Tested this on a Debian 12 which worked fine. One notice in that graph from `check_nrpe` against `ntp.se` that the offset is more fluctuating than with `ntpd`. Still within acceptable margin but an unexpected outcome.
> 
> Screenshot 2024-10-22 at 09 39 05

Yeah one possible outcome of might be that "ntp.se" used by the check and the node selected for tracking by chrony (see `chronyc tracking`) is not the same i guess, not sure really. At least it is well within acceptable standards for NTP, and now we are also using `NTS` on top of it so it is harder to spoof time.

PR #210, opened 2024-10-22
Enable account linking again

theseal:
Was disabled in 4.2.6 as stated in the comment


mickenordin (approved):
Should be safe since we know swamid will give us trustworthy emails

PR #213, opened 2024-11-01
certbot::sync: Do not overwrite /etc/letsencrypt

eest:
Instead of having client machines overwrite their local /etc/letsencrypt
directory with the contents of the same directory on the sync server
instead sync to a local /opt/certbot/letsencrypt directory.

A drawback of this is that deploy hooks or configuration files will need
to operate on certs under /opt/certbot/letsencrypt/live instead of
/etc/letsencrypt/live but this means a sync client machine can also
install certbot locally for fetching e.g. certs for its own name using
acme without the local certbot and sync job stepping on each other.

This also means sync clients will need to drop their own sync-related
hook scripts into our custom /opt/certbot/renewal-hooks/deploy directory
instead of expecting them to be synced from the server machine. This has
the added feature that a sync server no longer needs to have hooks
loaded locally to spread them to client machines (and in cases where the
sync server is only serving certs the hooks do not need to check if they
are running on the correct machine before doing stuff).


theseal:
Did a few more improvment we talk about on Slack.
* Better handling of directories
* Stop templating "executables"

PR #215, opened 2024-11-08
Naemon needs help reaping it's defunct children

theseal:
Recommendation from naemon devs.

https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/


mickenordin (approved):
Perfect

mikaelfrykholm (approved):
LGTM.
https://docs.docker.com/reference/compose-file/services/#init
for reference.

PR #217, opened 2024-11-11
Make IPv6 SNAT rules match more of ULA range

eest:
When creating containers via compose using a network with "enable_ipv6: true" the IPv6 network will be selected out of the ULA range: https://docs.docker.com/engine/daemon/ipv6/#dynamic-ipv6-subnet-allocation https://en.wikipedia.org/wiki/Unique_local_address

This means the fd0c::/16 match is not wide enough. As an example I got the address `"Subnet": "fd6b:795b:49eb::/64"` in a lab. Just be prepred to NAT the whole fd00::/8 instead.


mikaelfrykholm (approved):
LGTM

mickenordin (approved):
Great!

PR #220, opened 2024-11-21
Prevent realm spoofing

theseal:


theseal:
The realm magic happens in the docker image (https://github.com/SUNET/docker-geteduroam-freeradius/commit/02c4c59d4d80728f5b6ad68d6facaccd7852fd81) this configuration just enables another virtual server which executes the check.

SpaceFarmer (approved):
LGTM

PR #238, opened 2025-01-14
Update script for checking validity of infra cert

mariahaider:
Sending output of the openssl commands to /dev/null makes the error messages proper. Otherwise the output can look messy. For example when the check becomes critical, it prints out as below.

```
Certificate will expire
Certificate will not expire
CRITICAL: notAfter=Sep 21 07:20:03 2024 GMT
```

Sending the output to 'null' will print out only the 'echoed' messages.

`CRITICAL: notAfter=Sep 21 07:20:03 2024 GMT`


theseal (approved):
LGTM

mickenordin (approved):
Perfect!

PR #239, opened 2025-01-16
Add maxmemory_policy configuration option.

masv3971:
maxmemory-policy handles redis/redics ability to garbage collect keys with ended TTL.

### DIFF
--- a/manifests/redictcluster.pp
+++ b/manifests/redictcluster.pp
@@ -8,6 +8,7 @@ class sunet::redictcluster(
   Optional[Boolean] $prevent_reboot = false,
   Optional[String] $image = 'registry.redict.io/redict',
   Optional[String] $tag = '7-bookworm',
+  Optional[String] $maxmemory_policy = 'noeviction',
 
 )
 {
--- a/manifests/redictcluster.pp
+++ b/manifests/redictcluster.pp
@@ -8,8 +8,8 @@ class sunet::redictcluster(
   Optional[Boolean] $prevent_reboot = false,
   Optional[String] $image = 'registry.redict.io/redict',
   Optional[String] $tag = '7-bookworm',
-  Optional[String] $maxmemory_policy = 'noeviction',
-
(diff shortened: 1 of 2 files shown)


manifests/redictcluster.pp
```diff
@@ -8,6 +8,7 @@
   Optional[Boolean] $prevent_reboot = false,
   Optional[String] $image = 'registry.redict.io/redict',
   Optional[String] $tag = '7-bookworm',
+  Optional[String] $maxmemory_policy = 'noeviction',
```
mickenordin, line 12: Maybe:
```
Optional[String] $maxmem = undef,
String $maxmemory_policy = 'noeviction',
```

templates/redictcluster/server.conf.erb
```diff
@@ -19,6 +19,7 @@ tls-ca-cert-file /etc/ssl/certs/infra.crt
 <% else %>
 port <%= @redictportnum %>
 <% end %>
+maxmemory-policy <%= @maxmemory_policy %>
```
mickenordin, line 22: and here:
```
<% if @maxmem %>
maxmem "<%= @maxmem %>"
maxmemory_policy <%= @maxmemory_policy %>
<% end %>
```

mickenordin (changes requested):
I wonder if this should be connected to actually setting `maxmem` https://redis.io/docs/latest/develop/reference/eviction/#maxmem 

See inline comments for how I am thinking about it

manifests/redictcluster.pp
masv3971: how about `Optional[Integer] $maxmem = 0`, is no limit and it is the default for 64-bit systems?

templates/redictcluster/server.conf.erb
masv3971: Don't want to force a `maxmem` limit since that's not necessarily an issue for a particular setup.

mickenordin (approved):

PR #240, opened 2025-01-17
adding ability to set uptime check values in hiera

richir-sunet:
Lookup value from hiera, otherwise use set parameters, otherwise use defaults from parameters. Also doing some cleanup and pointing to new class for old class.

Is `check_uptime_warning` and `check_uptime_critical` reasonable names for hiera?


mickenordin (approved):
LGTM

PR #246, opened 2025-01-27
We have jobs which are not run every week

theseal:
If the cleanup job removes output files from jobs run less then every week we get alert from the monitor.

### DIFF
--- a/files/scriptherder/cleanup_scriptherder
+++ b/files/scriptherder/cleanup_scriptherder
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+import re
+import os
+
+with os.popen("/usr/bin/crontab -l") as pipe:
+    for line in pipe:
+
+        if re.match("#", line):
+            # Skip comments
+            continue
+
+        regex = "/usr/local/bin/scriptherder"
+        if not re.search(regex, line):
+            # Skip everything not monitored by scriptherder
+            continue
+
+        tokens = line.split()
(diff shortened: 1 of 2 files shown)


mickenordin (approved):
LGTM

mikaelfrykholm (changes requested):
This looks scary and prone to errors. 
```# m h  dom mon dow   command```
```@yearly /usr/local/bin/scriptherder apa```
 
```
❯ ./cleanup_scriptherder
 Traceback (most recent call last):
   File "/tmp/./cleanup_scriptherder", line 63, in <module>
     job_name = tokens[position].replace("-", "_")
                ~~~~~~^^^^^^^^^^
 IndexError: list index out of range
```

theseal:
> This looks scary and prone to errors. 
I would agree if we were editing the crontab by our self but in this case it's puppet doing all the magic, consistency. 

`# m h dom mon dow command` `@yearly /usr/local/bin/scriptherder apa`
> 
> ```
> ❯ ./cleanup_scriptherder
>  Traceback (most recent call last):
>    File "/tmp/./cleanup_scriptherder", line 63, in <module>
>      job_name = tokens[position].replace("-", "_")
>                 ~~~~~~^^^^^^^^^^
>  IndexError: list index out of range
> ```

Can try to add some error handling so we don't crash but the only "bad" thing if this script doesn't match a requested scriptherder job is that might get a disk event after a while.

PR #252, opened 2025-02-05
Support for dockerhost2

theseal:
Worth notice is that `dockerhost2` operates without the network namespaces from `dockerhost` so the tooling around mdqp might need to be tweaked. That's nothing this PR can handle since the switch is done in ops-repo.


btmattsson (approved):
Ser OK ut vad jag kan se :-)

PR #253, opened 2025-02-13
s3 backup of docker registry

acrn:
Requires `registry_s3backup_hostname` and `registry_s3backup_bucket` to be configured, and the hiera key `registry_s3backup_config` must contain:

    [default]
    access_key = [...]
    secret_key = [...]

### DIFF
--- a/manifests/docker_registry2.pp
+++ b/manifests/docker_registry2.pp
@@ -1,11 +1,13 @@
 # host docker registry with compose
 class sunet::docker_registry2 (
-  String  $registry_public_hostname   = 'docker.example.com',
-  String  $interface                  = 'ens3',
-  Array $resolvers                    = [],
-  String  $registry_tag               = '2',
-  String  $registry_cleanup_basedir   = '/usr/local/bin/clean-registry',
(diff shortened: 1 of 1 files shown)


mickenordin (approved):
I think this is fine, so ok by me, but in the future it might be a good idea to look into duplicity, that can push backups to s3, and will not only make sure that we have a copy of our data, but allows us to do point in time recovery and use encryption for the backups. Using duplicity rather than a sync will help against crypto lockers also.

acrn:
> I think this is fine, so ok by me, but in the future it might be a good idea to look into duplicity, that can push backups to s3, and will not only make sure that we have a copy of our data, but allows us to do point in time recovery and use encryption for the backups. Using duplicity rather than a sync will help against crypto lockers also.

@mikaelfrykholm and I talked about ransomware protection this morning, and that is certainly something we'll have to deal with at some point. He had some experience with proxmox backup server, so that could also be an option.  S3 was mostly chosen for this because the old registry uses that for backups, but in a different way.

theseal:
Coming late to discussion but I think it would be good idea to evaluate backup through `baas2` as complement to s3.

PR #254, opened 2025-02-14
Kano scriptherder cronjob fix

mickenordin:
Add support for jobs that run more often than hourly
    
I don't now how much logs we should save for these jobs,
but drive has a satosa that runs 109 jobs every five minutes
and run time is above a minute with 14 days, and about 10 seconds
with 2 days, so I decided to hard code that for these types of jobs.
    
It also refactors the code to reduce cyclomatic complexity
adding two functions decide_days_to_save and decide_interval

It also fixes a logical bug


theseal (approved):
LGTM

PR #258, opened 2025-02-24
Don't use public hostname to resolve certificate file locations

acrn:
For the test instance both the public and private hostname should be `docker-test.platform.sunet.se`.

For the production instance they should be `docker.platform.sunet.se`, and when it goes live `registry_public_hostname` should be changed to `docker.sunet.se`

The corresponding variable in docker_registry.pp is called "fqdn"


mikaelfrykholm (approved):
LGTM

PR #267, opened 2025-03-19
fix rsyslog firewall syntax

SpaceFarmer:
This is to make the class compatible with modern servers running NFT. 
Using this manifest to make it backwards compatible with legacy servers running UFW:
- https://github.com/SUNET/puppet-sunet/blob/main/manifests/misc/ufw_allow.pp

This was the error encountered on a ubuntu-24 server using this class:
`Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Unknown resource type: 'ufw::allow' (file: /etc/puppet/cosmos-modules/sunet/manifests/rsyslog.pp, line: 64, column: 9) on node log-test-sto3-1.komreg.net`

### DIFF
--- a/manifests/rsyslog.pp
+++ b/manifests/rsyslog.pp
@@ -63,7 +63,6 @@ class sunet::rsyslog(
     if ($udp_port) {
         sunet::misc::ufw_allow { "allow-syslog-udp-${udp_port}":
           from  => $udp_client,
-          ip    => 'any',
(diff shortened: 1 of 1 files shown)


SpaceFarmer:
Tested on an older ubuntu20 server running UFW and it resulted in no change on the server.

PR #270, opened 2025-03-27
Mifr naemon fixes

mikaelfrykholm:
* Bump to new versions of grafana components. This enables to send otel data directly to loki. Which gives us labels for free. 
* Include kubernetes dashboards
* Be a bit more helpful by setting all admin users to grafana admins also.


theseal (approved):
LGTM

I did not review the json files!

PR #272, opened 2025-03-31
Legacy cruft

theseal:
The arrow seem to interfer with ensure_resource.
Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Illegal relationship operand, can not form a relationship with an Undef Value. A Catalog type is required. (file: /etc/puppet/cosmos-modules/sunet/manifests/pypi.pp, line: 27, column: 5) on node pypi-1.sunet.se

### DIFF
--- a/manifests/pypi.pp
+++ b/manifests/pypi.pp
@@ -24,32 +24,32 @@ class sunet::pypi (
       recurse => false,
     }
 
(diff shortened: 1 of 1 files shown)


SpaceFarmer (approved):
LGTM

PR #274, opened 2025-03-31
Init nftables, normaly done in sunet::server

SpaceFarmer:
When 'sshd_config: false' is set in sunet::server, nftables is not initiated. So we get this error:
`Error: Could not find resource 'Service[nftables]' in parameter 'notify' (file: /etc/puppet/cosmos-modules/sunet/manifests/dockerhost2.pp, line: 40)`

So the fix is to add it to the pypi class.

This is tested in production and works.


theseal (approved):
LGTM

PR #275, opened 2025-03-31
Enable dovadm replicator commands

delhage:


templates/mail/dovecot/dovecot.erb.conf
```diff
@@ -139,7 +136,23 @@ service aggregator {
     user = mail
   }
 }
+
+service replicator {
+  client_limit = 0
+  drop_priv_before_exec = no
+  process_limit = 1
+  process_min_avail = 0
```
mickenordin, line 144: The docs say that process_min_avail should be set to one, so it can start replicating immediately, did you set it to 0 for a good reason? https://doc.dovecot.org/2.3/configuration_manual/replication/

delhage: No, I don't have a good reason. I have reverted that.

mickenordin (approved):
LGTM

PR #280, opened 2025-04-03
add kolab plugins

delhage:
- **Bump shib plugin version**
- **Add config**
- **Add kolab roundcube plugins**


mickenordin (approved):
This approach should work with version `1.6.9-apache-2` of the roundcube image

PR #286, opened 2025-04-10
Cosmos is not run from crontab

theseal:


theseal:
> Looks good for cleaning up output files from scriptherder for `run-cosmos` job. But it doesn't remove outputs for cronjobs that run once a day. Logic for `tokens[0] != "*":` is missing.

Lines starting with a digit like
```
32 * * * * /usr/local/bin/scriptherder --mode wrap --syslog --name dehydrated_fetch_acme-c.sunet.se -- sh -c 'ssh -Ti $HOME/.ssh/id_acme_c root@acme-c.sunet.se | /bin/tar xvf - -C /etc/dehydrated/certs && /usr/bin/le-ssl-compat.sh'
```
Are handled by the daily routine:
https://github.com/SUNET/puppet-sunet/blob/564f8e4b5e385533619d1fb8e38ef58395561373/files/scriptherder/cleanup_scriptherder#L26-L27

mariahaider (approved):
working as it should for daily cronjobs without the word @daily

PR #287, opened 2025-04-10
Add acme-d support for naemon

SpaceFarmer:
This PR adds support for running naemon/thruk with acme-d (in addition to acme-c that is default).

The use case for this is to allow for a server to not have to expose web ports towards the internet just for the sake of acme challenges on cert renewal.

In addition to this, the ability to configure nftables to allow only trusted network prefixes as source has been added to the class.

This feature branch is tested on a new deploy'ed server running acme-d, as well as an older one running acme-c.
No diff was detected on the older instance (as expected).


theseal (approved):
LGTM

PR #291, opened 2025-04-15
making list of load balancers a hiera variable

mariahaider:
Prevously the array of FQDN of load balancers was hardcoded. Now it is turned into a varaible with hiera lookup.


theseal (approved):
LGTM

PR #298, opened 2025-05-07
Haproxy stats visibility with IP range limitation

mariahaider:
Like the title says

### DIFF
--- a/manifests/lb/load_balancer/website.pp
+++ b/manifests/lb/load_balancer/website.pp
@@ -135,6 +135,7 @@ define sunet::lb::load_balancer::website(
 
   # Parameters used in frontend/docker-compose_template.erb
   $allow_ports            = pick_default($config['allow_ports'], [])
+  $stats_port             = pick_default($config['stats_port'], '')
   $dns                    = pick_default($config['dns'], [])
   $exposed_ports          = pick_default($config['exposed_ports'], [80, 443])
   $frontendtools_image    = pick($config['frontendtools_image'], 'docker.sunet.se/frontend/frontend-tools')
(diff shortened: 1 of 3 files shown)


manifests/lb/load_balancer/website.pp
```diff
@@ -228,6 +229,16 @@
     $frontend_ips_v4 = sunet::format_nft_set('', filter($frontend_ips) | $this | { is_ipaddr($this, 4) })
     $frontend_ips_v6 = sunet::format_nft_set('', filter($frontend_ips) | $this | { is_ipaddr($this, 6) })
     $external_interface = pick($config['external_interface'], $::facts['interface_default'], $interface)
+
+    if $config['stats_port'] {
+      $stats_dport = sunet::format_nft_set('dport', pick($config['stats_port'], []))
```
theseal, line 234: Why don't you reuse the variable set on line 138? I see that previous examples do the same but shouldn't it be enough with the first variable?
This second assignment uses another default then one line 138. Should it be a string or array?

theseal (changes requested):

manifests/lb/load_balancer/website.pp
mariahaider:
```
if $config['stats_port'] {
      $stats_port = sunet::format_nft_set('dport', pick($config['stats_port'], []))
```
Do you mean like above? I think I tried that, puppet didn't like it. I don't exactly remember the error message.

Regarding the second question, it should be string, i am fixing it, thank you for noticing it.

theseal: On line 138
https://github.com/SUNET/puppet-sunet/blob/2094db7fd99016aa971f030084d0bd0f27e10e0a/manifests/lb/load_balancer/website.pp#L138 we declare the `stats_port` variable.

Can't it be used like this:
```
if $stats_port {
      $stats_dport = sunet::format_nft_set('dport', $stats_port)
}
```

theseal (approved):
LGTM

PR #302, opened 2025-05-15
Workaround for https://codeberg.org/forgejo/forgejo/issues/7854

mikaelfrykholm:
All credit should go to @eest for doing the root-cause analysis.


acrn (approved):
works for naemon_monitor so it should solve our problem until fixed upstream

PR #304, opened 2025-05-19
The NRPE user wasn't allowed to read the compose

theseal:


mickenordin (approved):
Looks like a very smart person came up with this solution :stuck_out_tongue_winking_eye:

theseal:
> Looks like a very smart person came up with this solution 😜

Simply the best!

PR #307, opened 2025-06-04
Disallow web crawlers on forgejo hosts

acrn:
AI companies periodically fetch .tar.gz and .bundle files of all repositores. These files are cached on the server and use several Gbs of disk space, and the usage is duplicated with each backup.


theseal (approved):
LGTM

But does the evil AI bots respect robots.txt? Guess time will tell.

acrn:
> But does the evil AI bots respect robots.txt? Guess time will tell.

I can see that OpenAI and Claude have at least asked for the file, so hopefully they will respect it.

PR #315, opened 2025-08-11
Make sure the set the initial role for new users.

mikaelfrykholm:
We need to change the org_user entry as well for a user to get admin permission. Otherwise the user will manually need to powerclick their own permissions.


theseal (approved):
Powerclicking is no good. LGTM

PR #317, opened 2025-08-13
Systemd has been the default since at least Ubuntu 16.04 and Debian 9

theseal:
Also the fact is broken since the link for /sbin/init is no longer part of any package.


eest (approved):
LGTM

PR #323, opened 2025-08-19
baas2: improve argparse usage

eest:
Now the tool requires either --register or --install. Makes it more clear how to use the tool.


SpaceFarmer (approved):
LGTM

PR #324, opened 2025-08-19
IPv6 fix in dockerhost2

mariahaider:
Fix for enabling IPv6 by docker
Sources:

https://rexum.space/p/enabling-ipv6-support-for-docker-and-docker-compose/?cmdf=docker+compose+ipv6

https://github.com/moby/moby/pull/43197


mariahaider:
Yes, it is to enable Ipv6 per default. But the user can set this boolean value https://github.com/SUNET/puppet-sunet/blob/6f6d760624fa4cca2dd74801505e86e300de047d/manifests/dockerhost2.pp#L11 to false if one doesn't want to have ipv6 in the docker container. @mikaelfrykholm

PR #326, opened 2025-08-21
Mifr naemon fixes2

mikaelfrykholm:
Earlier I changed this role to use the native opentelemetry ingestion into loki. I missed that the logs did not get labeled correctly in this new flow, this fixes that. 
I also bumped the stack to more recent versions. 

The current thinking is that we want to align as much as possible with the opentelemetry semantics. I have changed the alloy senders out on the servers to send logs marked according to opentelementry semantics.


SpaceFarmer (changes requested):
As far as i can tell it looks good!
My only addition is that we might want to make sure that sqlite3 is installed on the server through `naemon_monitor.pp` since we execute commands that require it.

PR #327, opened 2025-08-22
Pahol otel opt-out

SpaceFarmer:
Add option to opt-out from configuring alloy. 

This can be useful if you need to exclude one server from sending telemetry after enabling it globally on a repo. For example the monitoring server since it uses the same alloy ports on the container as the installed alloy package causing a container reboot loop.


mikaelfrykholm (approved):
LGTM!

PR #330, opened 2025-09-05
Bump grafana key

mikaelfrykholm:
Bump grafana key. Used by otel::alloy and naemon_monitor.


theseal:
For traceability: https://grafana.com/blog/2023/08/24/grafana-security-update-gpg-signing-key-rotation/

PR #333, opened 2025-09-09
Update chrony config for Debian 13

eest:
On Debian 13 the default chrony config has been modified to use leapseclist instead of leapsectz so lets do that for our Debian 13 machines as well. Once we start seeing Ubuntu machines using leapseclist we can add a version check for Ubuntu as well.

Also Debian 13 has moved the confdir directive to the end of the config to fix https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1073865 and this feels like something we can do on both old and new machines.

Finally fix comment wording based on Debian 13 config so the diff is smaller on modern machines.

### DIFF
--- a/manifests/chrony.pp
+++ b/manifests/chrony.pp
@@ -23,6 +23,15 @@ class sunet::chrony(
   Array[String] $ntsserverkeys = [],
 ) {
 
+  $distro = $facts['os']['distro']['id']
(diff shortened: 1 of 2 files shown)


theseal (approved):
LTGM

mikaelfrykholm (approved):
LGTM!

PR #365, opened 2025-11-12
Mifr ceph needed fixes

mikaelfrykholm:
Allow dhcpv6 and don't discard multicast and broadcast as stuff breaks


mickenordin (commented):
Do we want to hide this behind a setting, so as to not interfere with current deployments?

PR #380, opened 2025-12-15
Thorslund edusigntest sp disco url

tuxdepend:
- Loadbalancers is now handled in ops-repo instead of static variable. 
- sp env variables are moved into an array variable to reduce duplication.
- disco url is changed for edugain


kushaldas (approved):
This looks good. Tested in test infra.

PR #401, opened 2026-01-27
Allow access to grub boot menu

theseal:
Default in the cloud images is to disable grub with `GRUB_TIMEOUT=0`. This is no good when needing to single user boot a machine in need.

Solution tested on
* Debian 11
* Debian 13
* Ubuntu 22
* Ubuntu 24

Test executed in IAAS but I see no risk or problem applying this configuration to all our servers.


mickenordin (approved):
👍

mikaelfrykholm (approved):
LGTM

PR #402, opened 2026-02-02
Opt-in notify at reboot

theseal:


templates/notify-reboot/sunet-notify-reboot.erb
mikaelfrykholm, line 1: Is it a problem that the url is publicy known?

mikaelfrykholm (approved):

templates/notify-reboot/sunet-notify-reboot.erb
theseal: Yes. The script kiddies might spam us that why we store it in Hiera.

PR #403, opened 2026-02-10
Mastondon thinning of caches

theseal:


files/mastodon/www/thinning.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+/usr/local/bin/tootctl media remove --days 7 --verbose
```
mickenordin: Maybe we should make these hard coded values configurable via a parameter?

mickenordin (approved):
Looks good, think about if we should make the values configurable. If you don't feel like it, that is ok too.

PR #405, opened 2026-02-16
"One job" now supports automount

theseal:
Added in v12.6.4.

https://code.forgejo.org/forgejo/runner/pulls/1231


mikaelfrykholm (approved):
LGTM

PR #407, opened 2026-02-18
Argus server provisioning from puppet-sunet

paulopamplona:
Argus (https://github.com/Uninett/Argus) is an alarm aggregator and notification tool that is been implemented mostly for CNaaS. 
There is an interest that other projects can make use of this to aggregate different nagios/naemon instances from other separate ops repos. 
This is the first draft for being able to create and provision argus using sunet::argus::server: .

### DIFF
--- a/manifests/argus/server.pp
+++ b/manifests/argus/server.pp
@@ -48,7 +48,7 @@ class sunet::argus::server (
     }
 
     if $argus_clients != '' {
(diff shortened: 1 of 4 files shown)


SpaceFarmer (approved):
Since this is a new class that is not used by anyone i see no risk with this PR.
My only feedback is to try and fix the puppet-lint warnings highlighted in this PR by code scanning engine.

PR #408, opened 2026-02-19
Improve performance

theseal:
This will not solve bad IO for disk but will hopefully improve other things.


mikaelfrykholm (approved):
LGTM

PR #409, opened 2026-02-20
Risk for lost logs

theseal:
It's not recommended to logrotate the docker logs if the json-file driver is in use (which we enforce in daemon.json). The driver will handle the rotation and the extra rotation by logrotate might cause lost logs (from `docker logs`, the logs in syslog are still safe).


eest (approved):
Related upstream issue: https://github.com/moby/moby/issues/45728
Specifically the truncation seems to be what is causing the problem and this is known upstream.

LGTM

PR #412, opened 2026-02-25
wget metatadata timeout and retires

mickets1:
Lowering default wget timeout and retries in order to avoid OOM on drive IDP servers when retrieving metadata.


theseal (approved):
I don't mind the change per se but I think we need to have a separate meeting regarding the metadata handling in Drive.

PR #415, opened 2026-03-10
Teach Alloy about pinned versions and config-dir for all config

mikaelfrykholm:


manifests/otel/alloy.pp
```diff
@@ -14,7 +15,7 @@
       unless  => 'dpkg -l alloy',
     }
     package { 'alloy':
-      ensure => 'installed',
```
theseal, line 17: Will old releases be available in the repo? 
I think that `apt::pin` is a better solution so that other mechanisms don't happen to upgrade the package

mikaelfrykholm:
They are. But you might be right. It is generally non ergonomic to select a certain version of a package as an apt upgrade  might want to upgrade it.

PR #416, opened 2026-03-17
nrpe checks for monitoring validity of both app and customer certs

mariahaider:
What the title says :)


theseal (changes requested):
Can you run `shellcheck` and `shfmt` on the new shell scripts and fix the outcome?

theseal (approved):
LGTM

PR #419, opened 2026-03-27
Fix acme-d certs for alloy

SpaceFarmer:
This needs to be done since the files in the 'live' directory is symlinks to the 'archive' directory. Else the alloy container will not have access to the letsencrypt certs.


mariahaider:
Looks good to me

mariahaider (approved):

PR #420, opened 2026-03-27
Use HTTPs for patroni API since the class now enforces it

SpaceFarmer:
We have switched to use https now for the patroni API and need to update the monitoring script accordingly.


theseal (approved):
LGTM

PR #423, opened 2026-04-02
Workaround since bad cert in classpath

theseal:
Caused by: java.io.IOException: Invalid PEM format:  No EOL char found in footer:  0 x0a

The problem was not the footer but the header


mariahaider (approved):
LGTM

PR #424, opened 2026-04-07
Forgejo runner 2.0 

theseal:


theseal:
This PR changed a bit since Forgejo 15 was released. Now we can do *real* ephemeral runners which not share the token of previous executions.

mikaelfrykholm (commented):
LGTM
great work!

PR #425, opened 2026-04-09
Secure patroni api

theseal:


SpaceFarmer (approved):
Secure API sounds good!
LGTM!

PR #427, opened 2026-04-14
New valkey class

SpaceFarmer:
New class to configure a valkey 9.x+ cluster.
It defaults to using the new infra CA with automatic cert renewals.


theseal (approved):
LGTM

PR #437, opened 2026-04-20
Mariadb modernize

theseal:
Don't mind the backup in branch name, it's just where I began…

Handle both old (pre 11.01) and modern images of mariadb.


mariahaider (approved):
LGTM

PR #438, opened 2026-04-21
datatype changes for allow_prefixes_by_tag

mickets1:
- Updated allow_prefixes_by_tag
- Added missing IPv4/IPv6 prefixes
- introduced NRPE check for validation
- Added Python script to verify host → prefix + tag mapping
- Fixed permissions so nagios user can execute the check, using sudoer.pp class
- Ensured all matching prefixes require correct tag (ex acmec)
- General cleanup


theseal (changes requested):
Please fix the puppet-lint stuff

theseal (approved):
LGTM

PR #439, opened 2026-04-24
alloy-unattended-upgrades

SpaceFarmer:
Add option to enable unattended upgrades for alloy


SpaceFarmer:
Turns out this was already handled in `sunet::server` if you use the option `unattended_upgrades_use_template: true`. So no need to reinvent the wheel.

PR #441, opened 2026-04-28
More modernization

mariahaider:
- Introduced new variables for opting boostrapping, monitoring, innodb_buffer_pool_size & healthcheck
- Removed variable - BOOSTRAP - since it is specific fo maraidb images built by Nextcloud
- Added built in NRPE check for checking cluster health, size etc

Documentaiton https://wiki.sunet.se/spaces/sunetops/pages/314212754/Mariadb


theseal (changes requested):
Please run shellcheck and shfmt on the shell script.

files/mariadb/check_galera_cluster
```diff
@@ -3,11 +3,10 @@
 #It is a changed version of the original https://github.com/fridim/nagios-plugin-check_galera_cluster/blob/master/check_galera_cluster
 
 
-PROGNAME=`basename $0`
+PROGNAME="basename $0"
```
theseal: Probably not a big problem but this change will remove the path as previously. Will just set a strange hardcoded name.
I think `$(basename $0)` is what you are looking for.

theseal (changes requested):

files/mariadb/check_galera_cluster
mariahaider: If I change as you suggested, it will complain about not having double quote. I can change it to `PROGNAME=$(basename "$0")`
```
Line 6:
PROGNAME=$(basename $0)
                    ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.
```
It worked with `PROGNAME="basename $0"` as well.

theseal (approved):
LGTM

PR #442, opened 2026-04-30
Raise the warning age to work with cron interval.

SpaceFarmer:
The cronjob runs every 15 minutes so 600 seconds is to low and will result in flapping alarms.


theseal (approved):
Äntligen!

mariahaider (approved):
LGTM2

PR #447, opened 2026-05-08
Fix issue running ubuntu26 on an ubuntu24 KVM.

SpaceFarmer:
`ubuntu26.04` or `ubunturesolute` was not in the `virt-install --osinfo list` on a ubuntu24 KVM server. 
This resulted in the following error:
`"ERROR    Unknown OS name 'ubunturesolute'. See --osinfo list for valid values."`

Fixing this by setting a fallback value of `ubuntu-lts-latest` which is always in the list. It's possible that there will be an update later to this list in libvirt that will contain ubuntu26.04 (and then it should not enter the fall-back any more).


theseal (approved):
LGTM

PR #449, opened 2026-05-12
valkey-fixes

SpaceFarmer:
- Fix broken OS logic in if-statement
- Fix bootstrap script so it uses the installed CLI tool
- refactor the certbot renewal hook so it works with the fact that the valkey server process inside the container runs with user `valkey` with id `999`. Without this the key file was not readable for the valkey process.
https://github.com/valkey-io/valkey-container/blob/19d0cbdd77fee11a548b74f85697e5973d4ef850/9.0/alpine/Dockerfile#L83-L85
`/data # cat /proc/$(pidof valkey-server)/status | grep Ui
Uid:	999	999	999	999`

### DIFF
--- a/files/valkey/certbot-renewal-hook
+++ b/files/valkey/certbot-renewal-hook
@@ -1,17 +1,55 @@
-#!/bin/sh -eu
-# vim: ai ts=4 sts=4 et sw=4
+#!/bin/sh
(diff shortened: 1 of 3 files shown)


mariahaider (approved):
LGTM

PR #452, opened 2026-05-19
Execute hooks even on first issuing

theseal:
Make it possible disable reload of the service which at the issuing state might not be fully functional.


SpaceFarmer (approved):
LGTM!

PR #453, opened 2026-05-19
Reusable class for SAML discovery

theseal:
Using thiss.io.


btmattsson (approved):
Looks good for me

templates/metadata/docker-compose-discovery.yml.erb
```diff
@@ -0,0 +1,46 @@
...
+<% if !@ds -%>
+    ports:
+      - "80:3000"
+<% end -%>
+    volumes:
+      - /opt/discovery/metadata/:/etc/thiss/:ro
+    environment:
+      - METADATA=/etc/thiss/metadata.json
+      - TRUSTINFO=/etc/thiss/metadata_sp.json
+      - BASE_URL=<%= @base_url %>/entities
```
mariahaider: We don't seem to have `base_url `anymore in mdq, i don't remember the reason.

templates/metadata/docker-compose-discovery.yml.erb
```diff
@@ -0,0 +1,46 @@
...
+    container_name: ds
+    image: <%= @ds_image %>:<%= @ds_version %>
+    environment:
+      - BASE_URL=<%= @base_url %>/
+      - MDQ_URL=<%= @final_mdq_search_url %>
+      - SEARCH_URL=<%= @final_mdq_search_url %>
+      - STORAGE_DOMAIN=<%= @domain %>
+      - LOGLEVEL=warn
+      - DEFAULT_CONTEXT=<%= @context %>
+<% if @mdq -%>
```
mariahaider, line 29: This part won't work if mdq is chosen but host port for mdq is not chosen to be configured. A specific variable for mdq host port can be chosen.

templates/metadata/get_ds_metadata.sh.erb
mariahaider, line 1: We added some more checks on this script after we had incidents where en empty metadata file was sent by aggregataor and docker container failed to start.

https://github.com/TheIdentitySelector/thiss-ops/blob/master/global/overlay/etc/puppet/modules/thiss/templates/mdq/get_metadata_prod.erb

theseal: Not sure what you mean. I would say that both scripts checks for the same thing. What are we missing?

templates/metadata/docker-compose-discovery.yml.erb
theseal: This is for the local mdq only which will be used inside the docker compose context by thissjs.

templates/metadata/docker-compose-discovery.yml.erb
theseal: Maybe @btmattsson can answer that

btmattsson: Might be since we are running mdq behind ds we needed that.

theseal: We will try update the class to not include the BASE_URL for the MDQ.

theseal: Tried without it and it works as intended 👍

templates/metadata/get_ds_metadata.sh.erb
theseal: Since we run the new script with `set -e` we don't need to check the return value from `wget`. If `wget` failes the script will terminate immediately.

mariahaider (approved):

PR #455, opened 2026-05-21
added ipmi class

tuxdepend:


SpaceFarmer (changes requested):
- `$password_max_lengt` should be `password_max_length`
- Don think you need the `path    => ` declarations when you have full path in the commands.
- the `set_ipmi_password` function could be wrapped with `sensitive` to prevent the password from ending up in any puppet logs. And the `onlyif  => "test -n '${admin_pass}'",` is redundant since you already check that the variable is set before hitting this function right? Maybe something like this:
`exec { 'set_ipmi_password':
  command   => Sensitive("/usr/bin/ipmitool user set password ${admin_user_id} '${admin_pass.unwrap}' ${password_max_length}"),
  logoutput => false,
  unless  => 'test -f /etc/ipmi_password_is_set',
  notify  => File['ipmi_password_is_set'],
}`

SpaceFarmer (approved):
LGTM!

PR #456, opened 2026-05-21
Designate ASN for load balancer servers

mariahaider:
We needed a solution to set ASN for the load balancer, the default is to have the same ASN as the route server.

An example for group.yaml:

```
    peers:
      sw1.sunet.se:
        as: '4200020051'
        local_as: '4200020061'
        remote_ip: 'x.x.x.x'
      sw2.sunet.se:
        as: '4200020051'
        local_as: '4200020061'
        remote_ip: 'x.x.x.x'

```


eest (approved):
LGTM

Could potentially use `''` instead of `undef` to look more like existing `Optional[String]` arguments, but i would not consider it a blocker.

PR #462, opened 2026-06-02
disable_resolved_stub: reload systemd-resolved when stub config changes

fredrikt:
Changing the `/etc/resolv.conf` symlink alone does not cause
systemd-resolved to re-read its configuration. Add
`systemctl reload-or-restart systemd-resolved` to both exec
commands so the stub enable/disable actually takes effect.


theseal:
Thanks

PR #463, opened 2026-06-02
check_docker_containers: delimiter/case-agnostic container matching

fredrikt:
## Summary

- Works with both Docker Compose v1 (underscores, original case) and v2 (dashes, lowercase), covering both `sunet::dockerhost` and `sunet::dockerhost2`
- Replaces per-container `docker inspect` calls with a single `docker ps --all` + one batched `docker inspect` on all IDs
- Matching uses `_normalise_containername()` which strips replica/run suffix, lowercases, and normalises `_` to `-` before comparing — no puppet-side configuration needed
- Reduces docker calls from up to ~5 per container to exactly 2 total regardless of container count

### DIFF
--- a/manifests/dockerhost.pp
+++ b/manifests/dockerhost.pp
@@ -19,8 +19,6 @@ class sunet::dockerhost(
   Boolean $advanced_network                   = false,
 ) {
 
-  $container_name_delimiter = '_'
-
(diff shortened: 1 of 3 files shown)


theseal:
Great!

PR #464, opened 2026-06-03
Mikand acmec changes

mickets1:
dehydrated_one
- Removed obsolete code, added error handling and safeguards.
- Force cert renewal for one domain

dehydrated_wrapper
- Added lockfile handling, timeout if process get stuck to allow continued script execution.
- adding timeouts for ORDER_TIMEOUT and VALIDATION_TIMEOUT. As dehydrated doesn't set defaults the process might get stuck.

remove_unused_checks
- Now included in dehydrated_wrapper instead of a separate cronjob
- removes old unused checks/domains not existing in local.yaml, including cached scriptherder information
- removing exclusions array and changing cleanup job name in order to not having to maintain an exclusions list.

### DIFF
--- a/files/dehydrated/remove_unused_checks.py
+++ b/files/dehydrated/remove_unused_checks.py
@@ -1,11 +1,18 @@
 import yaml
 import os
+import sys
 
 # Remove acmec checks and cached check data for domains not existing in local.yaml
 def remove_unused_checks(all_acmec_domains):
+    has_errors = False
     exclusions = ["gather_inventory", "dehydrated", "update_and_upgrade", "cleanup", "check_infra_cert", "cosmos"]
-    files = os.listdir("/etc/scriptherder/check/")
-    cfiles = os.listdir("/var/cache/scriptherder")
+    
+    try:
+        files = os.listdir("/etc/scriptherder/check/")
+        cfiles = os.listdir("/var/cache/scriptherder")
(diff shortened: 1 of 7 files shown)


mariahaider (changes requested):
The removal of scriptherder check can be kept in a separate script that we can run with a monthly cronjob. It needs to remove also the output files under '/var/cache/scriptherder/' for that specific domain.

mariahaider (approved):
Approved :)

files/dehydrated/remove_unused_checks.py
```diff
@@ -0,0 +1,74 @@
+import yaml
+import os
+import sys
+
+# Remove acmec checks and cached check data for domains not existing in local.yaml
+def remove_unused_checks(all_acmec_domains):
+    has_errors = False
+    exclusions = ["gather_inventory", "dehydrated", "update_and_upgrade", "cleanup", "check_infra_cert", "cosmos"]
```
theseal: I think this approach is looking to the problem in the wrong direction which will requires us to maintain this list of exclusions when a new job is added. Not ideal.

Isn't all the jobs prefixed with `dehydrated_`? If so, use that as a filter to only operate on what _we_ created already.

theseal (changes requested):

theseal (approved):

PR #465, opened 2026-06-03
dockerhost: Ubuntu 24.04+ apt key, nsrunc runtime, compose-plugin on 26.04

fredrikt:
Tested on freshly installed Ubuntu 22.04 and 26.04.

## Summary

- **Ubuntu 24.04+**: `apt-key` is gone; switch to `/etc/apt/keyrings/docker.asc` with the `keyring:` parameter in `apt::source`. The old `apt-key` path is retained as `elsif` for Ubuntu <= 22.04 and Debian.
- **nsrunc runtime**: Add a wrapper script that runs `runc` inside Docker's mount namespace. systemd 253+ (Ubuntu 24.04+) creates a private mount namespace for Docker; without this, container operations fail. Registered as the default runtime in `daemon.json`. Only written/registered when `write_daemon_config=true`, which is forced automatically on Ubuntu 26.04+ with a warning.
- **docker-compose-plugin**: Install the plugin package on Ubuntu 26.04+ instead of the legacy wrapper script.

### DIFF
--- a/manifests/dockerhost.pp
+++ b/manifests/dockerhost.pp
@@ -59,7 +59,28 @@ class sunet::dockerhost(
     }
   }
 
-  if versioncmp($facts['os']['release']['full'], '22.04') <= 0 or $facts['os']['name'] == 'Debian' {
+  # Ubuntu 24.04+ uses a new signed key format; apt-key is no longer available.
+  if $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['full'], '24.04') >= 0 {
+    apt::source {'docker_official': ensure => 'absent'}
+    file { '/etc/apt/keyrings':
+      ensure => directory,
+      mode   => '0755',
+    }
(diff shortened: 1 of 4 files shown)


manifests/dockerhost.pp
```diff
@@ -59,7 +59,28 @@
     }
   }
 
-  if versioncmp($facts['os']['release']['full'], '22.04') <= 0 or $facts['os']['name'] == 'Debian' {
```
theseal, line 62: I would rather see the use of `sunet::apt::repo_docker` as in `sunet::dockerhost2`.

theseal (changes requested):
We are really not using dockerhost non 2 any more but I guess we can merge this anyway of you fix my pending comment.

manifests/dockerhost.pp
fredrikt: Like this?

theseal:
Thanks

PR #466, opened 2026-06-03
docker_expose: allow_local param and unified dnat_dport

fredrikt:
## Summary

- Add `allow_local` param to permit intra-host Docker container access to the exposed service (adds nftables input + DNAT loopback rules)
- Add `dnat_dport` param that applies to both v4 and v6 DNAT, replacing the v6-only `dnat_v6_port`. The old param is kept as a deprecated alias with a warning.


theseal:
Thanks

PR #467, opened 2026-06-03
ssh_keyscan: AppArmor local override on Ubuntu 26.04+

fredrikt:
## Summary

Ubuntu 26.04 ships an AppArmor profile for `ssh-keyscan` that restricts it to network access only, causing the keyscan to fail when reading the hosts file. Add a local override at `/etc/apparmor.d/local/ssh-keyscan` to permit reading the hosts file, and reload the profile before running the keyscan.


theseal:
Great.

PR #468, opened 2026-06-03
fix File.exists? -> File.exist? in tls_certificates.rb (Ruby 3.2 compat)

fredrikt:
`File.exists?` was removed in Ruby 3.2. Replace with `File.exist?`.


theseal:
Thanks!

PR #471, opened 2026-06-08
apt/repo_docker: unify docker apt repo logic from dockerhost and dockerhost2

fredrikt:
I've only tested this on a freshly installed 22.04, because of issues with garethr-docker on 26.04. I will adress that in a separate PR.

It isn't exactly clear to me what this apt/repo_docker.pp is intended to do, so I'm guessing that it should handle setting up APT for Docker. It doesn't make sense to me to leave handling of older versions of APT in dockerhost.pp/dockerhost2.pp then, but I might have misunderstood.

I guess you haven't tested dockerhost2.pp on any 24.04? That version should require signed-by keyring rather than apt::key, and I suspect dockerhost2.pp was broken for 24.04.

## Summary

`sunet::apt::repo_docker` previously only handled Ubuntu 24.04+ / Debian 13+ (keyring-based apt). This PR extends it to also cover the older `apt-key` path (Ubuntu < 24.04, Debian < 13), moving that logic out of `sunet::dockerhost` and `sunet::dockerhost2`.

- `sunet::dockerhost` and `sunet::dockerhost2` both had duplicated inline apt repo setup for older distros. Both now unconditionally delegate to `ensure_resource('sunet::apt::repo_docker', ...)`, passing through `$docker_repo`.
- `sunet::apt::repo_docker` gains a `$docker_repo` parameter (default `stable`) and a version branch: older distros get the `apt-key`/`apt::source` path with the cosmos keyfile; newer distros get the `signed-by` keyring path.
- The pre-docker-ce package/source cleanup in `sunet::dockerhost` (lxc-docker, docker-engine, old sources.list.d entry, i386 arch removal) is dropped — these hosts are long gone.

### DIFF
--- a/manifests/apt/repo_docker.pp
+++ b/manifests/apt/repo_docker.pp
@@ -8,7 +8,7 @@ define sunet::apt::repo_docker (
   $release   = $facts['os']['distro']['release']['major']
 
   if ($distro == 'Ubuntu' and versioncmp($facts['os']['release']['full'], '24.04') < 0) or
-     ($distro == 'Debian' and versioncmp($release, '13') < 0) {
+    ($distro == 'Debian' and versioncmp($release, '13') < 0) {
     sunet::misc::create_dir { '/etc/cosmos/apt/keys': owner => 'root', group => 'root', mode => '0755'}
     file { '/etc/cosmos/apt/keys/docker_ce-8D81803C0EBFCD88.pub':
       ensure  => file,
--- a/manifests/apt/repo_docker.pp
+++ b/manifests/apt/repo_docker.pp
@@ -1,19 +1,55 @@
 # The apt repo for docker
 define sunet::apt::repo_docker (
+  Enum['stable', 'edge', 'test'] $docker_repo = 'stable',
 ) {
 
-  $distro = $facts['os']['distro']['id']
+  $distro    = $facts['os']['distro']['id']
   $lc_distro = downcase($distro)
+  $release   = $facts['os']['distro']['release']['major']
 
-  apt::source { 'docker':
-    location => "https://download.docker.com/linux/${lc_distro}",
-    repos    => 'stable',
-    key      => {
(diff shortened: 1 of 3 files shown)


theseal:
Sounds reasonably to handle everything regarding the APT config for docker in the separate class so I think you have understood it correctly. 

I don't do Ubuntu but @SpaceFarmer might have input in that matter.

fredrikt:
Any comments, @SpaceFarmer ?

SpaceFarmer:
Breaking out the APT related settings from the docker class seems like a good idea to me as well.
Regarding issues on ubuntu i have not noticed any on ubuntu-24, we have a lot of 24's running `dockerhost2:`. What i have seen is these kind of warnings on ubuntu-26:
`Warning: Target Packages (stable/binary-amd64/Packages) is configured multiple times in /etc/apt/sources.list.d/docker.list:3 and /etc/apt/sources.list.d/docker_ce.list:3
Warning: Target Packages (stable/binary-all/Packages) is configured multiple times in /etc/apt/sources.list.d/docker.list:3 and /etc/apt/sources.list.d/docker_ce.list:3
Warning: Target Translations (stable/i18n/Translation-en) is configured multiple times in /etc/apt/sources.list.d/docker.list:3 and /etc/apt/sources.list.d/docker_ce.list:3
Warning: Target DEP-11 (stable/dep11/Components-amd64.yml) is configured multiple times in /etc/apt/sources.list.d/docker.list:3 and /etc/apt/sources.list.d/docker_ce.list:3`

fredrikt:
Maybe I was misinformed about when the keyring change happened then. Glad it's working on 24.04 for you :)

I don't see those warnings on my Ubuntu 26.04 with this patch, but I'm also running on a modern puppetlabs-apt in case you are not.

Good to merge then?

SpaceFarmer:
Sounds good, i'll merge it to testing :)

PR #472, opened 2026-06-09
valkey tls improvements

SpaceFarmer:
Improve TLS related settings for valkey
- Announce the hostname instead of IPs to help with SAN validation from connecting clients
- Disable advertisement of non-TLS port
- Explicitly announce the cluster TLS port (used by the connecting clients)


mariahaider (commented):
looks good to me

PR #473, opened 2026-06-10
mariadb-tls

SpaceFarmer:
This PR makes it possible to configure a mariadb galera cluster with TLS enabled for both cluster traffic and client traffic.

It is optimised for use with the new infra CA with automatic cert reloads on letsencrypt/certbot renewals. (We have only been able to test that the the cert is reloaded for client connections, we have not found a way to check this for `wsrep` and `SST`, we just have to wait for the old certs to be sure(without restarting the cluster))

Implements and enforces encryption for:
- SST
- galera replication
- client connections
- backup replication

This is tested on mariadb version `12.3.2`

Also adding some other improvements:
- Make the purging of binlog retention configurable (6 hours is to low)
- Fix broken backup stream command that used the wrong binary
- Fix misconfiguration where `gtid_domain_id` and `wsrep_gtid_domain_id` was set to the same value
- Fix misconfiguration where `gtid_domain_id` was set to the same value on all nodes


mariahaider (approved):
:+1:

theseal:
MariaDB `11.4` has almost the same support date as `12.3`. Do you think it will work there as well?

files/mariadb/certbot-renewal-hook
```diff
@@ -0,0 +1,60 @@
...
+
+# === Sanity checks ===
+if [ -z "${RENEWED_LINEAGE:-}" ]; then
+	echo "ERROR: RENEWED_LINEAGE not set (not running from certbot?)" >&2
+	exit 1
+fi
+
+# === Ensure directory exists ===
+install -d -m 750 "${CERT_DIR}"
+
```
theseal, line 25: Why not use `install` with the correct permissions directly? No dancing with copy and move required.

theseal (changes requested):
Awesome pull request and feature but lets talk about the dancing with copy and move.

files/mariadb/certbot-renewal-hook
SpaceFarmer: The reasoning behind this was due to the bug in the install command in ubuntu-26 that does not allow you to set a user-id for a user that does not exist in the host OS. So the copy and move was to avoid having multiple code paths depending on OS.

SpaceFarmer:
> MariaDB `11.4` has almost the same support date as `12.3`. Do you think it will work there as well?

I have not seen anything while doing the research for this that has indicated that this is a new feature in `12.3`. So my best guess would be that it works in `11.4` as well. One thing worth noting is that when we enabled this for our existing clusters we needed to redo the bootstrap of the cluster again.

files/mariadb/certbot-renewal-hook
theseal: Damn you Ubuntu with crappy tools!

I wonder if it would feel better to wrap the dance in a if statement only for Ubuntu and let proper operating system skip the dance.

theseal (approved):
LGTM

PR #474, opened 2026-06-10
sunet-reinstall: wait for dehydrated instead of sleeping 150s

fredrikt:
Replace the hardcoded `sleep 150` in the sunet-reinstall cronjob with an active wait loop that polls for a running `dehydrated` process (up to 5 minutes, 10s intervals).

This avoids the race without adding unnecessary delay on hosts where dehydrated finishes quickly or isn't running at all.


theseal (approved):
Thanks

PR #475, opened 2026-06-10
server: dmi guards and skip grub on Raspbian

fredrikt:
Guard all dmi fact access with `facts.dig()` / `!facts['dmi']` checks so Puppet doesn't abort on nodes without DMI data.

Skip grub config entirely on Raspbian (no grub installed). Use `os.distro.id` instead of `os.name` for the Raspbian check — `os.name` reports `'Debian'` on Raspbian, causing the guard to be ineffective.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -83,7 +83,7 @@ class sunet::server (
     class { 'sunet::security::disable_all_local_users': }
   }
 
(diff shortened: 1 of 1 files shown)


theseal (approved):
Thanks. Perfect for all our Pi's 😉

PR #476, opened 2026-06-10
kvm: bootstrap cosmos via systemd service on second boot

fredrikt:
Three commits improving KVM cloud-init provisioning:

1. **Fix cloud-init schema warnings** — switch to `users` list with `passwd` key, `disable_root: false`, drop `package_update: true`.

2. **Bootstrap cosmos via systemd service** — write a `cosmos-bootstrap.service` unit via `write_files`, enable it, then reboot instead of running cosmos directly from `runcmd`. The reboot is needed in some situations (e.g. interface renames) and also fixes SSH server key restoration from `sunet-reinstall.tgz`. The service runs after `network-online.target` and removes the flag file on success, making bootstrapping idempotent — it retries on each reboot until finished.

3. **Rename NIC to eth0 via netplan** — replace the gateway6-conditional `60-disable_ra.yaml` with a permanent `60-cloud-init_user-data.yaml` that matches the NIC by MAC address and sets the name to `eth0`. The `accept-ra: false` setting is preserved inside the gateway6 conditional.

### DIFF
--- a/templates/kvm/user_data.erb
+++ b/templates/kvm/user_data.erb
@@ -2,11 +2,12 @@
 #
 # Docs: https://cloudinit.readthedocs.io/en/latest/ (this is Data source 'No Cloud')
 #
-user: root
+users:
+  - name: root
 <% if @password -%>
-password: <%= @password %>
+    passwd: <%= @password %>
 <% end -%>
-disable_root: 0
+disable_root: false
 chpasswd:
   expire: False
 ssh_pwauth: False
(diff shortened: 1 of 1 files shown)


theseal:
Haven't bootstrapped a KVM in a long time. Given the old way I would imagine that the output of the cosmos bootstrap would end up on the console? Will this change make it harder to debug boostrap issues?

fredrikt:
> Haven't bootstrapped a KVM in a long time. Given the old way I would imagine that the output of the cosmos bootstrap would end up on the console? Will this change make it harder to debug boostrap issues?

Console output is retained, and journald output is gained:

      StandardOutput=journal+console
      StandardError=journal+console

theseal:
Great, thanks.

PR #477, opened 2026-06-18
Making LBs new CA friendly

mariahaider:
The certificate materials are copied to old CA paths to adjust with the use of current puppet fact `tls_certificates`. This can be changed when we have new puppet facts matching the new CA.

### DIFF
--- a/manifests/lb/load_balancer.pp
+++ b/manifests/lb/load_balancer.pp
@@ -106,14 +106,17 @@ class sunet::lb::load_balancer(
     })
 
     each($websites) | $site, $config | {
-      if 'infra_certificate_bundle' in $config {
-        exec { 'reload_container':
-          command     => "systemctl restart frontend-${site}.service",
-          subscribe   => File['/opt/frontend/config/ssl/infra_haproxy.crt'],
-          refreshonly => true,
+      if 'haproxy_volumes' in $config {
+        each($config['haproxy_volumes']) |$volume| {
+          if $volume =~ /infra_haproxy.crt/ {
+            exec { 'reload_container':
+              command     => "systemctl restart frontend-${site}.service",
+              subscribe   => File['/opt/frontend/config/ssl/infra_haproxy.crt'],
+              refreshonly => true,
+            }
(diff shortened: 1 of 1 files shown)


manifests/lb/load_balancer.pp
```diff
@@ -72,6 +72,28 @@
...
+        owner  => 'root',
+        group  => 'root',
+        mode   => '0644',
+        links  => 'follow',
+    })
+    ensure_resource('file', $infra_key, {
+        source => "${certbot_dir}/privkey.pem",
+        links  => 'follow',
+    })
+    ensure_resource('file', $infra_bundle, {
```
theseal: Previous bundle contained both key, cert and chain. This will contain only the cert and chain. Is that as intended?

mariahaider: Thank for catching that descrepency. I have fixed it so that the bundle is created manually.

theseal (approved):
LGTM

johanlundberg (commented):
All hosts running lb:load_balancer needs to be setup for the new CA before this can be run? Or am I missing something?

mariahaider:
@johanlundberg All hosts running lb:load_balancer needs to be setup for the new CA if they use SUNET Infra certs. Right now, only SUNET load balancers uses SUNET Infra certs.

theseal:
> @johanlundberg All hosts running lb:load_balancer needs to be setup for the new CA if they use SUNET Infra certs. Right now, only SUNET load balancers uses SUNET Infra certs.

Should it be optional then?

mariahaider:
@theseal It is already optional. There is a logic that it will create the infra cert only if the puppet facts are found.
https://github.com/SUNET/puppet-sunet/blob/afae174e30ac55685500a478a353622cf5b4bfc7/manifests/lb/load_balancer.pp#L72

I just changed the inside of the logic so only new SUNET CA certs are used.

johanlundberg (approved):
LGTM

theseal (approved):
LGTM

PR #479, opened 2026-06-22
Fix exabgp fragments sort order with new puppet-concat

fredrikt:
The `exabgp_header` and `exabgp_footer` concat fragments had their order
values swapped (`header=10000`, `footer=10`), producing invalid config on
sites running a modern concat module (8+) which sorts numerically.

The fix uses zero-padded values (`'010'`, `'100'`, `'200'`) so that the
first character always differs. This makes lexicographic sort (old concat)
and numeric sort (concat 8+) agree on the correct order — both produce
header → neighbors → footer.

Tested on Ubuntu 22.04 (concat 2018, lexicographic sort) and Ubuntu 26.04
(concat 8.0.0, numeric sort).

### DIFF
--- a/manifests/exabgp/config.pp
+++ b/manifests/exabgp/config.pp
@@ -13,14 +13,14 @@ define sunet::exabgp::config(
 
   concat::fragment { 'exabgp_header':
     target  => $config,
-    order   => '10000',
+    order   => '010',
     content => template('sunet/exabgp/exabgp.conf_header.erb'),
     #notify  => Service['exabgp'],
   }
 
   concat::fragment { 'exabgp_footer':
     target  => $config,
-    order   => '10',
+    order   => '200',
     content => template('sunet/exabgp/exabgp.conf_footer.erb'),
     #notify  => Service['exabgp'],
   }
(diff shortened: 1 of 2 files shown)


theseal:
Adding @mariahaider for review (and merge if ok).

mariahaider:
I am aware of this 'reverse' concat in modern puppet versions. It does put the `process watch-service ` at the start of exabgp.conf, but it doesn't create an invalid configuration. It has been working in our load balancers running ubuntu 24 and debian 12. We have no example of running it on ubuntu 26. Does anything break there? @fredrikt 

The exaple here https://github.com/exa-networks/exabgp/wiki/Configuration-Syntax, does put the 'process' first.

We also added the api bit here https://github.com/SUNET/puppet-sunet/blob/testing/templates/lb/exabgp/exabgp.conf_neighbor.erb which mentions the process later in the configuraton.

Regarding the header file, we have no use of this https://github.com/SUNET/puppet-sunet/blob/testing/templates/lb/exabgp/exabgp.conf_header.erb, any longer since we run versions that are higher than 3.

Maybe, we should have a different name for the footer file instead and change the order to 090 for that file.

PR #480, opened 2026-06-22
ubuntu26 flatten fix

paulopamplona:
scope.flatten deprecated after puppet5

https://doc.wikimedia.org/puppet/puppet_functions_ruby3x/flatten.html
https://blog.x-way.org/Linux/2024/07/31/Puppet-ERB-array-flatten.html


theseal:
Works on older machines?

PR #481, opened 2026-06-23
Making (Simple) HAproxy compatible with New SUNET CA and dockerhost2

mariahaider:
https://github.com/SUNET/puppet-sunet/blob/testing/manifests/haproxy/simple_setup.pp is used for HAproxy instances in SUNET Auth Servers. The PR is to modernize it so new auth servers can run with dockerhost2 and new SUNET CA.

- Auth server's HAproxy configuration uses ACL to permit web requests from load balancers, since dockerhost2 makes it difficult to 'see' the acutal host IPs so needed to set specific IPs to the docker container running HAproxy and added 'special' DNAT nftables rules to facilitate the direct routing 
- added a variable to be able to choose CA cert path
- Configured preference for IPv4 in HAproxy configuration when it saves the DNS for backend (docker container running auth app) since the app's process doesn't listen to ipv6

### DIFF
--- a/manifests/apt/repo_docker.pp
+++ b/manifests/apt/repo_docker.pp
@@ -1,55 +1,20 @@
 # The apt repo for docker
 define sunet::apt::repo_docker (
-  Enum['stable', 'edge', 'test'] $docker_repo = 'stable',
 ) {
 
-  $distro    = $facts['os']['distro']['id']
+  $distro = $facts['os']['distro']['id']
   $lc_distro = downcase($distro)
-  $release   = $facts['os']['distro']['release']['major']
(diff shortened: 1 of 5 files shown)


theseal (changes requested):
Hmm. I think you need to refresh this PR with the latest `sunet::dockerhost2` and `sunet::apt::repo_docker` otherwise recent changed will be made undone.

johanlundberg (approved):
LGTM

mariahaider:
@theseal right, i needed to test without the errors, but now i know how to fix it. So, checked them out again from testing.

theseal (approved):
Great

PR #482, opened 2026-06-23
Add cleanup of old docker apt repo

SpaceFarmer:
We ran into this issue on existing servers:
`E: Conflicting values set for option Signed-By regarding source https://download.docker.com/linux/ubuntu/ noble: /etc/apt/keyrings/docker.asc != E: The list of sources could not be read.`
This is due to having 2 files for the docker repo: 
- `/etc/apt/sources.list.d/docker.list`
- `/etc/apt/sources.list.d/docker_ce.list`

So to fix this for existing servers we add a cleanup step.


mariahaider (approved):
LGTM

PR #483, opened 2026-06-23
dockerhost: replace garethr-docker class with direct resources

fredrikt:
## Summary

- Replace both `class {'docker': ...}` blocks in `sunet::dockerhost` with direct `file`, `exec`, and `service` resources, removing the dependency on the garethr-docker Puppet module
- Add `file { '/etc/systemd/system/docker.service.d': ensure => directory }` to replace the old `exec mkdir` hack needed to avoid duplicate declaration with garethr
- Add `file { '/etc/default/docker': ensure => absent }` to clean up the garethr-created file
- Add `service_override.conf.erb` systemd dropin for hosts not using `write_daemon_config` (passes DNS, storage driver, extra parameters, TCP bind, TLS via `ExecStart=` override instead of relying on garethr's service management)
- Reorder blocks to match `sunet::dockerhost2` (docker-compose → logrotate → docker-upgrade)

The approach mirrors how `sunet::dockerhost2` handles Docker service management.

### DIFF
--- a/manifests/dockerhost.pp
+++ b/manifests/dockerhost.pp
@@ -6,7 +6,8 @@ class sunet::dockerhost(
   $storage_driver                             = undef,
   $docker_extra_parameters                    = undef,
   Boolean $run_docker_cleanup                 = true,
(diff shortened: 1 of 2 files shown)


theseal:
Thanks.

PR #486, opened 2026-06-26
scriptherder cleanup: quote name to please find

johanlundberg:
example output before fix:

find: paths must precede expression:
`update_and_upgrade__ts-20260618T083001.902235_pid-3719781_output.data' find: possible unquoted pattern after predicate `-name'?


theseal (approved):
LGTM

PR #489, opened 2026-06-30
Make sure required resources exits

theseal:
Handle issue introduced in recent fix in dockerhost #483.

Error: Could not find resource 'Docker_network[docker]' in parameter 'require' (file: /etc/puppet/cosmos-modules/sunet/manifests/docker_run.pp, line: 101) on node dsmd-4.fidus.sunet.se


SpaceFarmer (approved):
LGTM

PR #491, opened 2026-06-30
scriptherder: random_sleep for cronjob, exclude param for monitoring

fredrikt:
## Summary

- `scriptherder::cronjob`: add `random_sleep` parameter to pass `--random-sleep N` to scriptherder, spreading cron job execution across a time window
- `scriptherder::monitoring`: add `exclude` array parameter to pass `--exclude` to the catch-all `check_scriptherder`/`check_scripts` NRPE commands; uses a canonical `$scriptherder_check_cmd` variable shared between the NRPE config and sudoers templates


theseal (approved):
LGTM

theseal:
Awesome that we don't have to implement random sleep in each job run by scriptherder! 👏

PR #492, opened 2026-06-30
Frontend improvements: route-adjust, docker compose v2, nftables, API enhancements

fredrikt:
## Summary

A collection of frontend improvements developed and tested on our local branch.
The commits are organised to be reviewed individually or adopted selectively.

### Infrastructure

- **docker_run: add `run_user` param** — lets callers specify uid:gid so containers run as non-root without hardcoding it in images.
- **docker_compose_service: Ubuntu 26.04+ uses docker compose v2 plugin** — detects Ubuntu 26.04+ and selects the compose2 systemd service template accordingly.

### Frontend networking / startup

- **configure-container-network: nftables-aware veth setup; docker compose v2 container naming** — handles both old (`instance_haproxy_1`) and new (`instance-haproxy-1`) compose v2 container naming; replaces the old bridge-based veth setup with a version that works under nftables (moves the veth into the docker network namespace where applicable, sets up source-routing for anycasted addresses).
- **start-frontend: fix zombie children after execv; support docker compose v2 plugin** — replaces `multiprocessing.Queue` with a pipe + double-fork so the intermediate child exits before the parent `execv`s into docker-compose (avoiding zombies); adds `--compose_bin /usr/bin/docker` support for the v2 plugin.

### Health-based VIP routing

- **frontend: health-based VIP routing via frontend-route-adjust** — replaces the old Python monitor script with a two-part design:
  - `exabgp_monitor` (bash): creates a FIFO, pushes initial route state, then relays ongoing updates from the FIFO to exabgp stdout.
  - `frontend-route-adjust@<instance>` (systemd path+service): watches the per-instance `announce` file written by the health-check container; on change, pushes route add/withdraw commands into the FIFO and manages kernel routes on `to_<instance>`.
  - `frontend-route-adjust-stop@<instance>` (oneshot): withdraws all routes cleanly on frontend stop.
  - exabgp and its FIFO dir (`/var/run/frontend-exabgp`) run as a dedicated non-root `exabgp` user; a `tmpfiles.d` entry recreates the dir on boot.

### API

- **frontend/api: allow_clients param, configurable port, run_user** — `api/server` gains `allow_clients` (nftables DNAT rule via `sunet::nftables::docker_expose`), a configurable `port` (default 8080), and `run_user` for non-root execution. `api/instance` gets a configurable `group` for the backends directory so the API container can write to it.

### Bug fixes / compatibility

- **frontend/docker-compose_template: haproxy_control path, HOSTFQDN Puppet 8 fix, STATSSOCKET env** — corrects the haproxy control socket path to `/haproxy_control`; fixes `HOSTFQDN` to use `$facts['networking']['fqdn']` (the `@fqdn` variable is not available in Puppet 8 templates); adds `STATSSOCKET` env var.
- **frontend: nftables, docker0-absent, Puppet 8 compat fixes** — `peer`: guard `is_ipaddr()` against undef `local_ip`; `load_balancer`/`website2`: use `facts.dig()` for `docker0` IP so nftables hosts (no docker0) don't crash; `website2`: replace unconditional `ufw route` exec with a conditional nftables conf.d rule.
- **websites2_monitor: poll-only fallback when inotify unavailable** — kernel 5.15.0-1100 returns ERRNO=0 from `inotify_add_watch`; catch `InotifyError` and fall back to a pure polling loop.

### DIFF
--- a/manifests/docker_compose_service.pp
+++ b/manifests/docker_compose_service.pp
@@ -15,8 +15,11 @@ define sunet::docker_compose_service(
     default => $service_name,
   }
 
-  $_template = $::facts['dockerhost2'] ? {
-    yes => 'sunet/dockerhost/compose2.service.erb',
+  $_use_compose2 = ($::facts['dockerhost2'] == 'yes') or
+    ($facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['full'], '26.04') >= 0)
+
+  $_template = $_use_compose2 ? {
+    true    => 'sunet/dockerhost/compose2.service.erb',
     default => 'sunet/dockerhost/compose.service.erb',
   }
 
--- a/manifests/docker_run.pp
+++ b/manifests/docker_run.pp
@@ -23,6 +23,7 @@ define sunet::docker_run(
   Hash[String, String] $extra_systemd_parameters = {},
   Boolean $uid_gid_consistency = true,
   Boolean $fetch_docker_image  = true,
+  Optional[String] $run_user   = undef,
 ) {
 
   $docker_class = $::facts['dockerhost2'] ? {
--- a/manifests/exabgp.pp
+++ b/manifests/exabgp.pp
@@ -1,10 +1,11 @@
 # exabgp
 define sunet::exabgp(
-  Array   $extra_arguments = [],
-  String  $config          = '/etc/bgp/exabgp.conf',
-  String  $version         = 'latest',
-  String  $volume          = '/etc/bgp',
-  Array   $docker_volumes  = [],
+  Array            $extra_arguments = [],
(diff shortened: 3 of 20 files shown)


mariahaider:
@fredrikt most of the changes except Health-based VIP routing & the zombie process fix in `start-frontend` script are already in https://github.com/SUNET/puppet-sunet/tree/stable-2023v1/manifests/lb. These are the puppet configurations we use in all Haproxy load balancing services in SUNET currently. We can anyway merge your changes under `manifests/frontend ` folder as well. I don't know which configurations you use in your system but do you have any plan to update the monitoring script for exabgp under `manifests/lb` as well or I can do it myself and test that it works in our system?

PR #494, opened 2026-07-07
fix apt get update hanging issue

naretev:
Fix an [issue](https://bugs.launchpad.net/ubuntu/+source/apt/+bug/2003851) where apt-get hangs indefinitely under certain conditions. This fix is for servers which are running a version of apt older than [3.1.3](https://launchpad.net/ubuntu/+source/apt/3.1.3)


acrn (commented):
Looks good.  Just add a comment about the versions after which this is no longer needed and it's good to go.

acrn (approved):
looks good

PR #496, opened 2026-07-16
Disable downloading of repository bundles

acrn:
Most LLMs seem to respect robots.txt, but from time to time an agent will download all tags of all repositories all formats offered.  These files are cached on the server and can use up hundreds of gigabytes of storage overnight until the server crashes.


theseal:
Sounds wise but I can't find any information regarding `ENABLE_ARCHIVE`. Maybe `DISABLE_DOWNLOAD_SOURCE_ARCHIVES` is enough?

PR #499, opened 2026-07-21
Don't expose all acmed certificate in certbot::sync

theseal:
And also make sure that the directory for hooks are always available for other classes to put files in


SpaceFarmer (approved):
LGTM

PR #500, opened 2026-07-23
Possibility to configure mimir more specificly

theseal:
E.g allow more request with `-querier.max-outstanding-requests-per-tenant=2048`


mikaelfrykholm (approved):
LGTM

PR #501, opened 2026-08-04
added dig function in redis.pp to handle missing interface keys

tuxdepend:
## Background

Node `db-dcoa-1.eduid.se` failed to compile its catalog with:

Error: Evaluation Error: Operator '[]' is not applicable to an Undef Value.
(file: manifests/redis/server.pp, line: 28, column: 5)

**Root cause:** commit `a89f94f6` *"Fix legacy facts"* replaced legacy top-scope facts
(`$::ipaddress_bond0`) with structured facts
(`$facts['networking']['interfaces']['bond0']['ip']`). Legacy facts degrade to `undef`
harmlessly; deep-indexing a structured fact is **not** undef-safe — if an intermediate key
(`bond0`) is missing, the trailing `['ip']` raises the error above.

`db-dcoa-1` has no `bond0` interface, so the lookup crashed. **Fixed** in `redis/server.pp`
using `$facts.dig(...)`.

## Other instances of the same pattern

A grep for interface-specific deep-indexing (`$facts['networking']['interfaces']['<if>']['ip']`)
found the following latent bugs. `$facts['os'][...]` lookups and
`has_key($facts['networking']['interfaces'], ...)` checks are **safe** and excluded.

| # | Location | Code | Risk |
|---|----------|------|------|
| 1 | `manifests/frontend/load_balancer/website2.pp:110` | `pick($facts['networking']['interfaces']['docker0']['ip'], $facts['networking']['ip'])` | **High** — `pick` fallback defeated; arg crashes before `pick` runs if `docker0` is missing |
| 2 | `manifests/frontend/load_balancer.pp:80` | `pick($facts['networking']['interfaces']['docker0']['ip'], 'no-address-provided')` | **High** — same `pick`-masked crash |
| 3 | `manifests/frontend/load_balancer/website.pp:68` | `$_docker_ip = $facts['networking']['interfaces']['docker0']['ip']` | **Medium** — only guarded by `sunet_nftables_enabled != 'yes'`, not by `docker0` existence |
| 4 | `manifests/lb/load_balancer/website.pp:69` | `$_docker_ip = $facts['networking']['interfaces']['docker0']['ip']` | **Medium** — identical to #3 |
| 5 | `manifests/bird.pp:13` | `undef => $facts['networking']['interfaces']['eth0']['ip']` | **Low** — only reached if `$router_id` is undef (defaults to `networking.ip`), then crashes if no `eth0` |

> **Note on #1 and #2:** the `pick(...)` was meant to provide a fallback, but the crash occurs
> during argument evaluation, so the fallback never protects anything. These fail on any host
> without a `docker0` interface (fresh install before Docker starts, or non-Docker hosts).

## Fix

Replace deep-indexing with `dig()`, which returns `undef` cleanly on a missing interface.

**Redis (applied):**

```puppet
$eth0_ip  = $facts.dig('networking', 'interfaces', 'eth0', 'ip')
$bond0_ip = $facts.dig('networking', 'interfaces', 'bond0', 'ip')

$slave_node = $master_ip ? {
  $eth0_ip                   => 'no',
  $bond0_ip                  => 'no',
  $facts['networking']['ip'] => 'no',
  default                    => 'yes',
}

pick cases (#1, #2):

$statsd_host = pick($facts.dig('networking', 'interfaces', 'docker0', 'ip'), $facts['networking']['ip'])

Direct-assignment cases (#3, #4, #5):

$_docker_ip = $facts.dig('networking', 'interfaces', 'docker0', 'ip')

### DIFF
--- a/manifests/redis/server.pp
+++ b/manifests/redis/server.pp
@@ -23,11 +23,15 @@ define sunet::redis::server(
   }
 
   # Configure all nodes as slaveof $master_ip, unless this node actually has $master_ip on eth0/bond0
+  # Use dig() so a missing interface (e.g. no bond0) yields undef instead of raising
+  # "Operator '[]' is not applicable to an Undef Value" from deep-indexing a structured fact.
+  $eth0_ip  = $facts.dig('networking', 'interfaces', 'eth0', 'ip')
+  $bond0_ip = $facts.dig('networking', 'interfaces', 'bond0', 'ip')
   $slave_node = $master_ip ? {
-    $facts['networking']['interfaces']['eth0']['ip']    => 'no',
-    $facts['networking']['interfaces']['bond0']['ip']   => 'no',
+    $eth0_ip                   => 'no',
+    $bond0_ip                  => 'no',
     $facts['networking']['ip'] => 'no',
-    default              => 'yes',
+    default                    => 'yes',
   }
 
   if $slave_node == 'yes' {

### CODE REVIEW

**Comment by mariahaider** [comment]
I think, docker0 is created if dockerhost2 is used. We do not use the manifests under `frontend` folder anymore, at least I am not aware of any use of those. And all the load balancers are supposed to use Docker and have `sunet_nftables_enabled` facts anyway.  It is an OLD setup like the comment says.
```
 if $::facts['sunet_nftables_enabled'] != 'yes' {
    # OLD setup
    $_docker_ip = $facts['networking']['interfaces']['docker0']['ip']
    # On old setups, containers can't reach IPv6 only ACME-C backend directly, but have to go through
    # a proxy process (always-https) running on the frontend host itself.
    $_letsencrypt_override_address = $facts['networking']['ip']
  } else {
    # NEW setup with Docker in namespace
    $_docker_ip = '172.16.0.2'  # TODO: Parameterise this somehow
    $_letsencrypt_override_address = undef
  }
```
But it is probably a good idea to check first that docker0  exists. You may open a PR for it.

Regarding `bird.pp`, I see that the default for `router_id` is `$facts['networking']['ip']`, don't know why we have the undef logic here, maybe it can be set as undef when calling the class if the user knows that the host actually has eth0 defined.

```
class sunet::bird(
  String $bird       = 'bird',
  String $bird6      = 'bird6',
  String $package    = 'bird-bgp',
  String $username   = 'bird',
  Integer $uid       = 501,
  Integer $gid       = 501,
  String $router_id  = $facts['networking']['ip'],
  String $check_args = '',
) {
  $my_router_id = $router_id ? {
    undef   => $facts['networking']['interfaces']['eth0']['ip'],
    default => $router_id
  }
```

Did you test this change for redis.pp in older OS like ubuntu 22 or 20?

tuxdepend:
dig() is supported from puppet 4.5
oldest server i could find ran 5.5.10 and this target server runs 5.5.22
So it is unlikely that this will cause backwards comparability problems.

mariahaider (approved):

PR #503, opened 2026-08-04
Facilitate TLS certficates generated by ACME DNS challenges in load balancers 

mariahaider:
We can use configuration for SUNETs certbot sync server with load balancers. 

https://github.com/SUNET/puppet-sunet/tree/stable-2023v1/manifests/certbot/sync
https://wiki.sunet.se/spaces/sunetops/pages/209485964/sunet+certbot+sync

We created puppet facts so that puppet clas for HAproxy load balancers can use certificates generated using ACME DNS challenges.


SpaceFarmer (approved):
LGTM

PR #504, opened 2026-08-04
Dehydrated cert cleanup for clients

mariahaider:
This can be useful to use in HAproxy load balancers where certificates from multiple domains are fetched.


manifests/dehydrated/client_cleanup.pp
```diff
@@ -0,0 +1,27 @@
+class sunet::dehydrated::client_cleanup(
```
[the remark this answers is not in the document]
theseal: Just add a space between the # and the first character will fix this issue I think.

manifests/dehydrated/client_cleanup.pp
```diff
@@ -0,0 +1,28 @@
...
+  $gc_conf = "${basedir}/dehydrated-cleanup.conf"
+
+  file { $gc_conf :
+    owner   => 'root',
+    mode    => '0640',
+    content => "BASEDIR=${basedir}\nWELLKNOWN=/var/tmp\n",
+  }
+
+  sunet::scriptherder::cronjob { 'dehydrated_cleanup':
+    cmd           => "sh -c 'test -x /usr/sbin/dehydrated && /usr/sbin/dehydrated --cleanup --config ${gc_conf}'",
```
theseal, line 23: How will dehydrated know which certs to archive?

theseal (commented):
Is your intention to run this on the loadbalancer where multiple stale domains and/or certs can be found?

manifests/dehydrated/client_cleanup.pp
mariahaider: It will archive all the certs under `BASEDIR`.

This is how the config file looks like

❯ cat /etc/dehydrated/dehydrated-cleanup.conf
BASEDIR=/etc/dehydrated
WELLKNOWN=/var/tmp`

mariahaider:
> Is your intention to run this on the loadbalancer where multiple stale domains and/or certs can be found?

Yes, but mostly for invalid old certs. We have also a use case where we backup the `/etc/dehydrated/certs `folder (https://github.com/SUNET/puppet-sunet/blob/e3d1b86e37360250a948c3d4ea79fe50d265578e/templates/kvm/sunet-reinstall.erb#L60) in a seed image, I think it is unnecessary to backup old certs.

I saw the cleanup script running in eduID load balancers so thought of integrating it to other load balancers through puppet-sunet.

theseal (approved):
LGTM

